PR #109, opened 2026-01-09
max's `project` code

nikloynes:
starting PR to give feedback/comments, add more commits


nikloynes (commented):
nice - a lot to get through there. i think there are some really interesting, inspired ideas here. the conceptualisation as a `project` makes sense (although i'm not sure this is fully ideated yet - e.g. is there an overlap betwen `project` and `run`?). i think the idea behind the CLI is really good, and the typer framework provides a surprisingly easy way of achieving this. i'd jig the CLI around a bit (see comments) but most of what's there can achieve that quite easily i think. 

i have one general point of criticism which is that this PR encompasses IMO way too much functionality, none of which is directly addressing/referencing an issue in our backlog. it seems to me that this PR covers: rejigging data model definitions, implementing batch multi-file processing, adding a CLI, defining/proscribing lots of new files, folders, formats, etc., and new conceptualisations of core modules for the app (project/run). 
every single one of these additions should ideally be at least one (if not several) self-contained PR(s), with unit tests. this big PR is hard to add to merge into the head branch of the project with good maintainability in mind. downstream, this means a lot of work for the team in either dis-entangling what of this PR can be used, or re-building components from within it from scratch, because they might actually not match the requirements. i think it would be good, for future contributions, to keep this in mind, and try to perhaps look at smaller chunks, specifically those that are prioritised in the project board. 

my second point is that a lot of the ideas here - while, again, are well thought out and pretty cool - are quite opinionated and give the user little flexibility in how they might want to use the software. usually, some kind of default approach will transpire while we build our application, but we don't necessarily want to hard-code that default, or force the user to use it. perhaps this is a matter of personal preference... 

in terms of next steps, perhaps you can reply to a few of the comments that ask for clarification, e.g. rationale for moving function a to module b, etc. i think then we can go through some of the functionality in here (specifically `project.py` data model and the typer scripts) and copy over some stuff into 2 new PRs that separate those features in order to get them into dev branch quickly.

harryjmoss (commented):
I think this could have the diff reduced significantly with a merge in of the latest changes to `development`, at least in terms of the references to `app` vs `deet` that I don't see in the head of `development`

PR #120, opened 2026-01-19
move processedAnnotationData to Base

### LINKED ISSUE #131: Inconsistent class inheritance in data models
### Description

We're not using TypeVars for Generic models as suggested in pydantic [docs](https://docs.pydantic.dev/2.4/concepts/models/#generic-models). As a consequence we're leaning on subtypes more than I think strictly necessary.

In `deet.data_models.eppi`, we have the idiosyncracy that `EppiGoldStandardAnnotation` inherits from `GoldStandardAnnotation` https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/935fc3b1667c14a9c797779c8dfb44c76d5a496b/deet/data_models/eppi.py#L234

But `EppiGoldStandardAnnotatedDocument` inherits from `EppiDocument`

https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/935fc3b1667c14a9c797779c8dfb44c76d5a496b/deet/data_models/eppi.py#L262

For consistency's sake it should inherit from `GoldStandardAnnotatedDocument`, but it can't because it would have to override the `documents` field.

This will cause a problem if we have a function that does something to either a GoldStandardAnnotatedDocument or an EppiGoldStandardAnnotatedDocument, e.g.

```python
def proc_annotated_document(document: GoldStandardAnnotatedDocument) -> None:
   print(document.document_id)
```

If we were subclassing correctly, both `proc_annotated_document(gold_standard_document)` and `proc_annotated_document(eppi_gold_standard_document)` would work, but as things stand mypy will complain.

An example of how this already creates problems is that `ProcessedAnnotationData` contains methods (e.g. `populate_custom_prompts()`) that could be applied to base Attributes. We should only write these once, but creating a `ProcessedEppiAnnotationData` that inherits from a base `ProcessedAnnotationData` is not straightforward, because as with `GoldStandardAnnotatedDocument`, we can't override the generic lists with eppi-specific ones https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/935fc3b1667c14a9c797779c8dfb44c76d5a496b/deet/data_models/eppi.py#L317-L330

There's a solution to this in https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/pull/120#issuecomment-3790345777, but it may cover all cases.

I think there are also more cases where EppiTypes are used when a BaseType would work equally well. It would be nice to get rid of these if we can and use Base types wherever possible

### Definition of Done

- [ ] All subtypes in eppi specialize generic base classes using TypeVar-parameterized models.
- [ ] Functions and methods written for eppi classes that can be generalised are generalised

### Comments

I've used the bug label, but it's not exactly a bug but an architectural issue

mcallaghan:
# Pull Request

## Description
Moves `ProcessedAnnotationData`, which was the result of parsed eppi.json, and contained general functions for parsed goldstandard annotation data (such as filtering attributes by csv) into base, and created an EppiSpecific function that subclasses this. 

To do this, we introduce TypeVars and Generics, to better manage inheritance from base classes to eppi. This is necessary because if a model `ProcessedAnnotationData` has a field, which is a list of another model (e.g. `documents: list[Document]`, this cannot safely be overwritten in a model that subclasses ProcessedAnnotationData.

This PR also fixes bidirectional inheritance problems, where GoldStandardAnnotatedDocument subclassed Document, and EppiDocument also subclassed Document. This meant that EppiGoldStandardAnnotated document could not follow the pattern of subclassing the corresponding base class:

```bash
        Document  ───────────────►  EppiDocument
           │                           │
           │                           │
           ▼                           ▼
GoldStandardAnnotatedDocument ─────► EppiGoldStandardAnnotatedDocument
```

This is achieved through a breaking change. 

GoldStandardAnnotatedDocument no longer inherits from Document. Instead, document becomes a field of GoldStandardAnnotatedDocument
 

## Related Issue(s)
Fixes #131 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Testing
`deet/scripts/pipeline_prompt_csv_01.py` and `deet/scripts/pipeline_prompt_csv_02.py` both still work.

## Additional Notes

---


nikloynes (commented):
looks good, few minor comments re fields that should be eppi-native versus not. slight concern about the functionality for CLI for interactive prompt population in eppi-specific data staying the same, so far that reason (and the fact it's not there yet) it will be good to test that functionality for both base and eppi csv/prompt population methods. 

might also be good to wait for #121 to be merged as it has significant refactor in `base/eppi.py` and might avoid some duplicate work. 

note also linting errors in CI.

nikloynes (commented):
tests are great, and everything apart from the `TypeVar` thing in base.py looks good to go, i think there's just a bit of conversation to be had here before we persevere with this.

PR #189, opened 2026-03-19
Add minimal CSV-to-DEET converterFeature/covidence import

IshitaGopal:
# Pull Request

## Description
This PR replaces the previously closed PR (#166).
I've incorporated the review feedback from that discussion and rebased the changes onto main.

### Summary of changes:
- Improved docstrings
- Updated CSV column type inference to use the full column as the default and ignore None values
- Added support for user-defined attribute and citation/reference fields, with sensible defaults when not provided
- Added support for creating destiny_sdk-compatible ReferenceFileInput and enhancement models

Please continue the review here.

### Next Steps
- Add support for sanitizing DOIs
- Document available destiny_sdk options
- Add support for user-defined columns to exclude
- Add support for correctly infering type for columns with multiple tags
- Add tests
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Pre-Review Checklist

- [] All existing and new tests pass locally and in CI

## Testing

Tests not added yet.

## Additional Notes

---


L-ENA (commented):
Hi,

Thanks for creating this data processor! It looks great. 

I ran another Covidence dataset through it, and added comments and suggestions in places where there may be opportunities to make things more generalisable :)

Lena

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,580 @@
...
+        )
+
+        reference_dict["authorship"] = (
+            self._build_authorship_list(reference_dict["authorship"])
+            if "authorship" in reference_dict
+            and isinstance(reference_dict["authorship"], str)
+            else None
+        )
+
+        bibliographic_data = BibliographicMetadataEnhancement(**reference_dict)
```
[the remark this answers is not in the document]
IshitaGopal: Addressed in the latest commit. 

The CSV reader reads all values as strings, and Pydantic fails on empty strings when expecting integers or None (e.g: for publication_year). To get around this I just added a line that converts empty strings to None in _build_reference_dict_from_row() func . Pydantic parsing should now work without errors.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,580 @@
...
+            csv_reader.fieldnames = colnames
+
+            # validate duplicates
+            dup_fields = self._find_duplicate_column_names(colnames)
+            if dup_fields:
+                msg = f"{len(dup_fields)} Duplicate fieldnames found: {dup_fields}"
+                raise ValueError(msg)
+
+            # validate required fields
+            meta_fields = {"name", "document_id"}
```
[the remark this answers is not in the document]
IshitaGopal: The `Document` class the converter inherits from requires the `name` field to be specified, so I am keeping the name column requirement for now.

deet/data_models/base.py
```diff
@@ -106,6 +106,28 @@ def to_json_type(self) -> JsonValue:
 DEFAULT_ATTRIBUTE_TYPE = AttributeType.BOOL
 
 
+class ContextType(StrEnum):
```
nikloynes: @IshitaGopal `ContextType` is in `documents.py`. make sure to incorporate upstream changes from main into your repo by running 
```bash
git fetch origin
git merge origin/main
```

deet/processors/base_converter.py
```diff
@@ -17,7 +17,7 @@
 )
 from deet.data_models.processed_gold_standard_annotations import ProcessedAnnotationData
 
-DEFAULT_BASE_OUTPUT_DIR = Path("tmp_parsed_eppi")
+DEFAULT_BASE_OUTPUT_DIR = Path("tmp_parsed")
```
nikloynes, line 20: might this break stuff elsewhere/backwards compatibility?

deet/processors/converter_register.py
```diff
@@ -14,14 +14,23 @@ class SupportedImportFormat(StrEnum):
...
+        if self == SupportedImportFormat.EPPI_JSON:
+            from deet.processors.eppi_annotation_converter import (
+                EppiAnnotationConverter,
+            )
 
-        mapping = {
-            SupportedImportFormat.EPPI_JSON: EppiAnnotationConverter(),
-        }
-        return mapping[self]
+            return EppiAnnotationConverter()
```
nikloynes: why switch from mapping to if-else? if we add new converters, there's more to do here to update them. or is there a strong reason not using the mapping?

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+    GoldStandardAnnotation,
+)
+from deet.data_models.documents import (
+    Document,
+    GoldStandardAnnotatedDocument,
+)
+from deet.data_models.processed_gold_standard_annotations import (
+    ProcessedAnnotationData,
+)
+from deet.logger import logger
```
nikloynes: here you should import `from loguru import logger`, the deet logger only runs at app level, e.g. when the CLI is run

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+    A class to convert raw CSV (e.g. Covidence) annotations
+    into structured Pydantic models.
+
+     This converter operates on flat CSV columns, infers field/column types,
+    and produces attributes, documents, and annotated document records.
+    """
+
+    def __init__(
+        self,
+        base_output_dir: str | Path | None = DEFAULT_BASE_OUTPUT_DIR,
```
nikloynes, line 89: re the backwards compatibility mentioned earlier, we could just add a new constant for csvs here, that wouldn't break compatibility for existing EPPI users.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+            annotated_documents_filename: Filename for annotated documents output
+            attribute_mapping_filename: Filename for attribute ID to label mapping
+
+        """
+        if base_output_dir is None:
+            logger.debug(
+                "`base_output_dir` set to None; "
+                "converting to empty string for compatibility."
+            )
+            base_output_dir = ""
```
nikloynes, line 115: what are the downstream effects of is? presumably everything is written to current working dir?

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+            Outfiles.DOCUMENTS: (
+                documents_filename,
+                TypeAdapter(list[Document]),
+            ),
+            Outfiles.ANNOTATED_DOCUMENTS: (
+                annotated_documents_filename,
+                TypeAdapter(list[GoldStandardAnnotatedDocument]),
+            ),
+            Outfiles.ATTRIBUTE_LABEL_MAPPING: (
+                attribute_mapping_filename,
```
nikloynes, line 136: would be nice to have some comments here that explain in a bit more depth what's going on here.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+    def _infer_type(value: str) -> type | None:
+        """
+        Given a string value infer its type.
+        Returns bool, int, float, or str depending on the value.
+        Empty or null-like values return None.
+        """
+        if value is None or str(value).strip() == "":
+            return None
+
+        v = str(value).strip().lower()
```
nikloynes, line 175: do we really need the normalize method if we're just using builtins in our code? will the method change? if not probably fine to ditch and stick with this here.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+            return bool
+        try:
+            float(v)
+            if v.lstrip("-+").isdigit():
+                return int
+            return float  # noqa: TRY300
+        except ValueError:
+            return str
+
+    # TODO: need to add list and dictionary handling
```
nikloynes: what specifically do you mean by this to do?

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+        if unique == {str}:
+            return AttributeType.STRING
+        if unique == {bool}:
+            return AttributeType.BOOL
+        if unique == {int}:
+            return AttributeType.INTEGER
+        if unique.issubset({float, int}) and bool not in unique:
+            return AttributeType.FLOAT
+        msg = "Multiple incompatible types found"
+        raise ValueError(msg)
```
nikloynes: maybe a custom error would be good here, for easier handling later on.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+            rows = [
+                {"id": "1", "age": "42", "gender": "F"},
+                {"id": "2", "age": "", "gender": "M"},
+                {"id": "3", "age": "36", "gender": ""},
+            ]
+            fieldnames = ["id", "age", "gender"]
+
+            converter = CovidenceAnnotationConverter()
+            converter._infer_column_types(rows, fieldnames, sample_size=2)
+            # Output: {"id": int, "age": int, "gender": str}
```
nikloynes, line 247: i really like that you've added these examples here, really helpful. good documentation and docstrings in any case.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+
+        """
+        n_rows = len(rows)
+        raw_column_types: defaultdict[str, list] = defaultdict(list)
```
nikloynes: what's actully in here? nothing? maybe add a comment or re-think the obj name

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+
+                row_num += 1
+
+        resolved_column_types: dict[str, AttributeType] = {}
+        for col_name, col_types in raw_column_types.items():
+            try:
+                resolved_column_types[col_name] = self._resolve_types(col_types)
+            except ValueError as e:
+                msg = f"Error inferring type for field '{col_name}': {e}"
+                raise ValueError(msg) from e
```
nikloynes: again, might benefit from its own error.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+
+        return resolved_column_types
+
+    @staticmethod
+    def _build_reference_dict_from_row(
+        mapping: dict[str, str],
+        row: dict[str, Any],
+    ) -> dict[str, Any]:
+        """
+        Build a nested dictionary of reference data from a CSV row.
```
nikloynes, line 285: 'reference' has a pretty specific meaning in destiny/deet - is this a destiny reference? probably best to be explicit about that if so, or rename if not.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+            "publication_venue": {"display_name": "Nature Climate Change"},
+            "abstract": "ABCDEFXXXXXX"
+            }
+
+        """
+        result: dict[str, Any] = {}
+
+        for key, column_name in mapping.items():
+            if column_name not in row:
+                continue  # or raise, depending on strictness
```
nikloynes: we could add an arg in the method like `strict=False` and raise here if `strict=True`.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+        Args:
+            authors: A semicolon-separated string of author names. eg:("Alice; Bob; Mo")
+
+        Returns:
+            List of `Authorship` objects representing each author with their position.
+
+        """
+        # Split on semicolons, strip whitespace, and remove any empty entries
+        clean_authors = [
+            author.strip() for author in authors.split(";") if author.strip()
```
nikloynes: is it *always* gonna be a semicolon? do we maybe want to add a few things it can attempt to split by? or perhaps have the user tell us in an optional arg?

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+        `BibliographicMetadataEnhancement` and `AbstractContentEnhancement`
+        objects, then wraps them in `EnhancementFileInput` objects, and finally
+        returns a `ReferenceFileInput`.
+
+        Args:
+            row: Dictionary representing a single CSV row.
+            mapping: Dictionary mapping nested keys (dot-separated strings) to CSV
+                 column names.
+            source: Optional string indicating the source of the data; defaults to
+                "DEET CSV converter"
```
nikloynes: super nitpicky - but we mainly refer to the app as `deet` not `DEET` given that's its CLI name, and it's probably good to strive for consistency in that.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,609 @@
...
+
+            rows: List of dictionaries representing all CSV rows.
+
+        Returns:
+            A tuple containing: Document and List[GoldStandardAnnotatedDocument]
+
+        """
+        attr_by_label = {a.attribute_label: a for a in attributes}
+
+        # ---  ---
```
nikloynes: this can go

deet/processors/csv_annotation_converter.py
nikloynes, line 1: really good work. i've got some minor comments here and there, but this looks very close to being good to go to me.

nikloynes (commented):
looks great. minor comments here and there, but nothing huge. once you've gone through them, we still need tests for all the functionality in here before we can merge this PR. i think you can broadly copy what's going on in the other annotation converter unit test files.

deet/processors/base_converter.py
IshitaGopal: Good point! I searched for usages of `tmp_parsed_eppi` and it only appears in logs from previous runs, not in the codebase or tests. I also ran the tests and didn’t see any failures, so this change shouldn’t affect backward compatibility.

deet/processors/converter_register.py
IshitaGopal: No strong reason for switching away from mapping - wanted to do lazy imports. I agree the mapping is better, especially as we add more converters. I switched this back in the latest commit.

mcallaghan:
Just added a line to make sure init_document_identity is called when creating a document, otherwise linking will fail downstream. After fixing that, I can confirm that it works nicely on my dataset!

deet/processors/csv_annotation_converter.py
IshitaGopal: Thank you! Glad these are helpful. Spent a lot of time on it :).

deet/processors/csv_annotation_converter.py
IshitaGopal: yes, everything will be written relative to the cwd.

deet/processors/csv_annotation_converter.py
IshitaGopal: This and the `__init__` section were copied from the `eppi_annotation_converter` module without modification.

I’ve added some comments and docstring details to clarify the initialized arguments, the behavior of the default output directory, and the purpose of the OUTFILE_LOADERS registry. (It maps output types to filenames and corresponding Pydantic TypeAdapters used for serialization and validation.)

deet/processors/csv_annotation_converter.py
IshitaGopal: I use this down in the loop to access attributes by labels.

deet/processors/csv_annotation_converter.py
IshitaGopal: Refactored code so this check happens while loading the csv instead of here.

deet/processors/csv_annotation_converter.py
IshitaGopal: I kept the delimiter as a semicolon for now to avoid adding extra configuration.
I also avoided splitting on other characters like commas, since they can appear within names (e.g., “Last, First”), which could lead to incorrect parsing.
If we see more variation in input formats, we can make this configurable in the future.

IshitaGopal:
Thank you for all the feedback! I’ve pushed updates addressing all comments and added some unit tests for the converter. Ready for another review 🙌

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,674 @@
...
+            return AttributeType.BOOL
+        if unique == {int}:
+            return AttributeType.INTEGER
+        if unique.issubset({float, int}) and bool not in unique:
+            return AttributeType.FLOAT
+
+        msg = f"Multiple incompatible types found: {unique}"
+        raise ColumnTypeInferenceError(msg)
+
+    def _infer_column_types(
```
nikloynes, line 217: i actually wonder, for this method -- are there not potentially columns which don't really represent an attribute? for instance, `id`, or some other kind of metadata.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,674 @@
...
+
+        -----------------------------------------------
+        USER GUIDE
+        -----------
+        The `mapping` argument defines how CSV columns map to Destiny SDK fields.
+
+        Format:
+            mapping = {
+                "<destiny_field>": "<csv_column_name>",
+            }
```
nikloynes, line 307: cool!

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,674 @@
...
+        Rules:
+            - Keys are dot-separated Destiny SDK field paths.
+            - Values are CSV column names.
+
+        SUPPORTED FIELDS
```
nikloynes, line 312: where in this method are we actually constraining to supported fields?

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,674 @@
...
+                row_num += 1
+
+        resolved_column_types: dict[str, AttributeType] = {}
+        for col_name, col_types in column_type_evidence.items():
+            resolved_column_types[col_name] = self._resolve_types(col_types)
+
+        return resolved_column_types
+
+    @staticmethod
+    def _build_destiny_reference_dict_from_row(
```
nikloynes, line 280: i might have missed this, or perhaps it's elsewhere, but do we have some code that makes a best guess at the mapping without the user providing it?

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,674 @@
...
+
+    # TODO: Allow users to provide seperator
+    def _build_destiny_authorship_list(self, authors: str) -> list[Authorship]:
+        """
+        Build a list of a `Authorship` objects  (as defined in destiny_sdk.enhancements)
+        from a string. The returned list is in the format expected by
+        `BibliographicMetadataEnhancement` for authorship.
+
+        Args:
+            authors: A semicolon-separated string of author names. eg:("Alice; Bob; Mo")
```
nikloynes, line 371: right, so the semicolons are the standard in covidence? should be quite easy to add the custom sep arg, as i imagine it'll become relevant quite quickly.

deet/processors/csv_annotation_converter.py
```diff
@@ -0,0 +1,674 @@
...
+                annotations=annotations,
+            )
+
+            annotated_documents.append(annotated_doc)
+
+        return documents, annotated_documents
+
+    def process_annotation_file(
+        self,
+        file_path: str | Path,
```
nikloynes, line 644: generally speaking, i don't think we support plain strings for filepaths as type hints throughout deet. obiously the user can pass them, as `pathlib` has compatibility for that. but it might make life easier for mypy etc if you want to leverage pathlib's nice features without allowing plain str as well.

tests/unit/test_csv_annotation_converter.py
nikloynes, line 1: maybe you can have a look at how we've done tests throughout the codebase, usually not with a class, but with every unit test being its own function. it would be great if the tests here can also be like that.

nikloynes: otherwise nice looking tests here.

nikloynes (commented):
looking good. 
as mentioned in teams, perhaps you can share a minimum working example of how i'd use this in e.g. ipython along with some data, so we can do some interactive testing of this code. 

some general to dos:
- a few of the comments are maybe a little superfluous and can be removed, especially if we've got rich docstrings. 
- adapt tests to the style we've used in existing tests (classes->functions)

pretty close to merging!

deet/processors/csv_annotation_converter.py
mcallaghan: This is the same as it is for the eppi converter, so I think this is fine for now, good to keep these in sync since they both inherit from the ABC

deet/processors/csv_annotation_converter.py
IshitaGopal: Added CSVParserConfig class to provide parser configuration seetings which now includes an author_separator argument.

deet/processors/csv_annotation_converter.py
IshitaGopal: Added code where if `auto_assign_reference_fields = True` in the `CSVParserConfig` class, the parser will try to match column names to the expected keys of reference field (`ALLOWED_REFERENCE_MAPPING_KEYS`).

deet/processors/csv_annotation_converter.py
IshitaGopal: We check for this when loading the csv (in `load_csv').

mcallaghan (approved):
This is looking great, thanks for refactoring the tests. We are good to go now IMO

PR #204, opened 2026-03-30
Interactive cli

### LINKED ISSUE #150: Consistent approach to file storage
### Description

We've been using input and output paths passed to pipeline scripts to manage file input and output. This makes transferring data between components of a pipeline more fiddly.  As things get more complicated, and we begin to have multiple pipeline runs, managing these files becomes harder for users. We need a standardised approach to saving and organising the artefacts produced by pipeline runs, that allows us to implement the [chunked/batched](https://destiny-evidence.github.io/evaluation-book/index-1/#chunked-evaluation-data) evaluation approach we intend to follow. This may be along the lines of #109, see outline at #84. 

This could be done with the filesystem, or we could use a sqlite database. We should consider that we will likely be storing this in a database if DEET becomes a hosted multi-user multi-project web service.

We should also consider 

### Definition of Done

- [ ] Some kind of project structure to allow users to manage multiple pipeline runs on the same task, only importing and parsing once
- [ ] Pipeline runs, and the artefacts they produced are saved in a standardised way, that allows for easy evaluation
- [ ] Clear definition and documentation of the concepts (Project, Batch, etc.) we need to organise this data

### LINKED ISSUE #184: Constrain CLI to a specific workflow via a "deet init" command
### Description
Currently, the CLI has modular commands that allow a user to perform different kinds of operations on different types of data (e.g. PDF parsing to md, EPPI JSON importing and attribute.csv generating, linking with CSV mapper generation, etc).

While powerful, this can be daunting and confusing. In particular, it can be difficult for a user to understand how they should proceed and what files need processing/modifying when. 

As an easy starting place, user's could be directed to use a "deep init" command that will interactively ask the user for:

1. A project name (creating a local folder with that name prefixed by "deet" in which all files will be stored and deet will run from)
2. An EPPI JSON file to parse/import attributes from (copying this to the newly created local directory)
3. A directory/path to a collection of PDFs for linking (copying this to directory of PDFs to the newly created local directory)
4. A .env file for LLM configuration (copying this to the newly created directory)

Once completed, the user would be within this new project directory and will be able to run the existing commands to process their data as they see fit.

The command could also end with a list of suggested commands the user could to run through a simple workflow (i.e.. like the README.md)

### Definition of Done

- [ ] "deet init" interactive command added to CLI
- [ ]  command prompts creation of new project directory named via user input
- [ ]  command prompts and copies user provided EPPI json file to project directory
- [ ]  command prompts and copies directory of PDFs to project directory
- [ ] command prompts and copies .env file to project directory

### Dependencies
Probably the containerisation issue/feature for simplicity.
- #Issue

mcallaghan:
# Pull Request

## Summary

This pull request makes a major update to the CLI to increase usability and pave the way for future development of deet.

### Deet Projects

The main novelty is the introduction of a deet project, which is a construct to organise deet input and output for a specific data extraction task.

Given that many deet steps only need to be run once for a data extraction task (parsing gold standard data, creating a link map, linking data, creating a prompt csv), the new cli introduces a command `deet project init` to streamline the process. On initialising a project, users define the location and type of dataset they want to use, and the location of their pdfs. A link map, prompt csv, and config template are created automatically. 

A project is initialised inside a directory, and when deet is run from inside that directory, the location of the dataset, prompt csv, and link map do not have to be defined each time an extraction task is run.

A project also creates an experiments directory within a project. Each time an experiment is run, a folder is created, with a structure defined in `deet.projects.ExperimentArtefacts` that standardises where and how the results and details of an experiment (the config used, the prompts used, the metrics, the output of the llm, the comparison of the llm to the gold standard data) are stored.

This makes it easier in the future to compare across experiments, and to replay and experimental config that has been validated in a new setting.

### Interactivity and usability in the CLI

The standardisation that a project brings reduces the number of things that need to be configured with command line arguments. To further make the CLI easier to use, some things that could previously be configured with command line arguments, or by editing yaml or env files, can be configured through interactive menus in the CLI. Everything can still be run in non-interactive mode.

The PR also introduces more structured terminal output to guide users through using the CLI. Important information can be presented inside styled boxes, which can be configured using markdown (see for example [deet/ui/terminal/templates/project/success](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/development/deet/ui/terminal/templates/project/success.md))

### Re-organisation of commands

The commands in the first version of the cli are in a flat structure. As the list of commands grows, this will become unwieldy. The new cli nests commands under subcommands, where commands doing similar things are grouped.

Concretely,, `deet project ...` comprises all commands to do with setting up and managing a project, while `deet experiments ...` comprises all commands to do with running (and later compring) experiments. `deet --help` shows the two top-level commands, and `deet project --help` describes and explains the subcommands to do with projects.

## Description

**tldr**: 
- Separate concerns of `DataExtractionSettings` and `DataExtractionConfig`
- Formalise deet project structure; 
- separate commands into 
  - deet project (setting up and configuring a project)
  - deet run (running experiments on a configured project)
- use wizards to replace (optionally) cli args
- mini-UI framework using rich console and .md templates

### Separate concerns of `DataExtractionSettings` and `DataExtractionConfig`

Several options were defined in both places, which made things a bit confusing. Now,

- `DataExtractionSettings` contains things which should be secret, and kept in a file that is not committed (.env)
- `DataExtractionConfig` contains things which should be transparent and reproducible, and kept in files that are commited (yaml files)

### Deet Project Structure

`deet.data_models.project.DeetProject` stores project configuration options. These are things that don't change within a project, currently:
- Project name
- Raw data source and type
- PDF directory

From this, after initiating a `DeetProject` object, `DeetProject.setup()` processes data, creates a prompt csv file, creates a link mapping file, exports a config template, creates necessary directories, and dumps the project to .yaml.

Paths to resources like the prompt csv, the link mapper, etc. are stored as properties.

#### Deet projects in the CLI

Deet projects can be loaded with `DeetProject.load()`: loading a deet project assumes that commands are run from a directory containing a valid project toml file.

Project loading is attempted before any command is run, via a typer.callback, and a project | None is accessible via `typer_context.obj.project`. The decorator `deet.scripts.typer_context.project_required` can be used for any command where a valid project is required to carry out the command. Decorating a command with this will mean that it fails gracefully if no project is found.

### Separate commands

The cli used to have a large number of flat commands. these are now split into `deet project` for commands to initialise and configure projects, and `deet experiments`, for commands to run data extraction experiments. In the future, commands to compare results across experiments will live here too.

All previously defined commands now print a deprecation warning with instructions on how to run the command in the new version.

### Wizards

CLI args were a little overwhelming. Many arguments are now no longer necessary since we have projects. We also use [InquirerPy](https://inquirerpy.readthedocs.io/en/latest/) to do nice prompting in the terminal. Inquirerpy has some nice features like file path completion, and validation.

In practice, we do not define the wizards manually, but use `deet.ui.terminal.wizards.run_model_wizard` to auto-generate a wizard from a pydantic model, and then create an instance of that model based on the user input. When we do this, any field that has `deet.data_models.ui_schema.UI` metadata will be prompted for, based on the information in the UI schema.

All CLI commands can be run in non-interactive commands

### Mini-UI framework

We re-organise the architecture of how we render things to the terminal somewhat, defined in `deet.ui.terminal`. This architecture leaves open the door for defining other UIs as modules of deet.ui. We make better use of [rich](https://github.com/textualize/rich) (which comes with typer), to have nicer, more structured output to terminal. A panel component is defined in `deet.ui.terminal.components`, and we define ui text and formatting in `.md` templates in `deet.ui.terminal.templates`. These templates are rendered with jinja, so can be dynamic. For example, the output of `deet`, will differ depending on whether it is called from a project dir or not.

## Related Issue(s)
Closes #184, closes #150

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## Pre-Review Checklist

- [ ] Core functionality still works locally (e.g. all pipeline scripts)

## Additional Notes

---


deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def process_data(self) -> ProcessedAnnotationData:
+        """Process the project's gold standard data."""
+        converter = self.gold_standard_data_format.get_annotation_converter()
+        return converter.process_annotation_file(self.gold_standard_data_path)
+
+
+@dataclass(frozen=True)
+class ExperimentArtefacts:
+    """Defines the structure of an data extraction experiment directory."""
```
VJonah: nit: typo in "an data..."

deet/scripts/commands/experiments.py
```diff
@@ -0,0 +1,134 @@
...
+    ] = None,
+) -> None:
+    """
+    Extract data from documents and evaluate.
+
+    Load gold standard annotation data, and use an LLM to extract data from the
+    documents in your dataset. Evaluate by comparing the results to the gold
+    standard data.
+    """
+    import yaml
```
VJonah, line 75: I'm assuming imports are here because there would be circular imports if they were at the top-level?

deet/ui/terminal/templates/project/success.md
VJonah, line 1: These markdown files are great!

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+            yaml.safe_dump(config.model_dump(mode="json"), sort_keys=False),
+            encoding="utf-8",
+        )
+
+    @classmethod
+    def load(cls, filename: Path = PROJECT_FILE) -> "DeetProject":
+        """Load a project from a toml file."""
+        import toml
+
+        data = toml.load(filename.open())
```
VJonah: Doesn't this open the file without closing it? Would `with filename.open() as f: ...` be better?

deet/extractors/cli_helpers.py
VJonah, line 139: Maybe a type hint issue. `prepare_documents` hints a `Sequence[Document]` return type. Could cause unclear failures downstream if `fail_with_message` is handled unexpectedly (e.g. mocked in tests)

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+    PROJECT = auto()
+    SYSTEM = auto()
+
+    def env_file(self) -> Path:
+        """Map option to env file."""
+        mapping = {
+            EnvironmentFile.PROJECT: Path(".env"),
+            EnvironmentFile.SYSTEM: Path.home() / ".deet" / ".env",
+        }
+        return mapping[self]
```
nikloynes: not sure this is the best way to implent this. .env files in general are meant for local development, not really production. so IMO it should be either ENV or local project config file (which can take the form of a dotenv file). dotenv is just a workaround for mocking real ENV vars, so we don't want to keep it from being able to access those system level ENV vars, but allow .env (root) or  config file, formatted as .env (proj)

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+    """
+
+    # Wizard fields, configurable by users
+    name: Annotated[
+        str,
+        UI(
+            help="Give your project a name. This will help you to identify it later",
+            valid="Must be at least 2 characters",
+        ),
+    ] = Field(..., description="The name of a deet project", min_length=2)
```
nikloynes, line 58: create a randomly generated name if user leaves empty. 
e.g.
```python
from random_word import RandomWords
r = RandomWords()
name = f'{r.get_random_word()}_{r.get_random_word()}'
```

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+    environment_file: Annotated[
+        EnvironmentFile,
+        UI(
+            help=(
+                "Where to store your API keys."
+                " Select system if you want to re-use credentials across deet projects,"
+                " or Project if you wish to use specific credentials for this project"
+            )
+        ),
+    ] = Field(default=EnvironmentFile.SYSTEM, description="Environment file")
```
nikloynes: see previous comment: i think this should be renamed, to allow to get env vars from env, .env and local .env (and it should be able to do a lot of this by default)

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+                "If you want to extract data from full texts, "
+                "choose a directory that contains your pdfs."
+                " You will have an opportunity to link this later"
+                " by running `deet link-documents-fulltexts`"
+            )
+        ),
+    ] = Field(None, description="Path to folder containing PDFs")
+
+    # Project metadata
+    created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
```
nikloynes, line 98: user override possible to give custom date?

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+
+    # Root is set during load() or defaults to cwd
+    _root: Path = PrivateAttr(default_factory=Path.cwd)
```
nikloynes, line 101: notsure  this is the best strategy for finding the proj repo root dir

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+
+    @property
+    def root(self) -> Path:
+        """Return project root."""
+        return self._root
+
+    # Computed paths - file and folder structure within project dir
+    @property
+    def experiments_dir(self) -> Path:
```
nikloynes, line 110: what's this?

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+        """Return path to linked documents folder."""
+        return self.root / "linked_documents"
+
+    @property
+    def config_path(self) -> Path:
+        """Return path to config file."""
+        return self.root / "default_extraction_config.yaml"
+
+    @property
+    def env_path(self) -> Path:
```
nikloynes: see above

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+
+    # Configuration and validation
+    model_config = ConfigDict(
+        json_encoders={Path: str},
+        extra="ignore",
+    )
+
+    @field_validator("gold_standard_data_path", mode="after")
+    @classmethod
+    def _abs_and_check_suffix(cls, value: Path) -> Path:
```
nikloynes, line 147: docstring

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+        abs_path = value.resolve()
+        if abs_path.suffix not in SUPPORTED_EXTENSIONS:
+            unsupported_ext = f"Unsupported extension, allowed: {SUPPORTED_EXTENSIONS}"
+            raise ValueError(unsupported_ext)
+        return abs_path
+
+    @field_validator("pdf_dir", mode="after")
+    @classmethod
+    def _process_pdf_dir(cls, value: Path) -> Path | None:
```
nikloynes, line 157: docstring - unclear what this is doing

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+        if value == "":
+            return None
+        return value.resolve()
+
+    def setup(self) -> None:
```
nikloynes, line 163: setup_project. unclear name might cause trouble later on

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+
+        self.export_config_template()
+        notify("Exported default config template", level=LogLevel.SUCCESS)
+
+        self.experiments_dir.mkdir(parents=True, exist_ok=True)
+        self.linked_documents_path.mkdir(parents=True, exist_ok=True)
+
+        self.dump_to_toml()
+
+    def dump_to_toml(self, target: Path = PROJECT_FILE) -> None:
```
nikloynes: so we've got yaml, toml and json files. does it have to be toml? can it just be a text file, or a json?

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+        """Write a minimal ``project.toml`` file to save project options."""
+        import toml
```
nikloynes: should probably go to the top unless is a super greedy import or rarely used method

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+        config = DataExtractionConfig()
+        self.config_path.write_text(
+            yaml.safe_dump(config.model_dump(mode="json"), sort_keys=False),
+            encoding="utf-8",
+        )
+
+    @classmethod
+    def load(cls, filename: Path = PROJECT_FILE) -> "DeetProject":
+        """Load a project from a toml file."""
+        import toml
```
nikloynes: see above

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+        """Check if project exists in current directory."""
+        return PROJECT_FILE.exists()
+
+    def process_data(self) -> ProcessedAnnotationData:
+        """Process the project's gold standard data."""
+        converter = self.gold_standard_data_format.get_annotation_converter()
+        return converter.process_annotation_file(self.gold_standard_data_path)
+
+
+@dataclass(frozen=True)
```
nikloynes, line 220: why not pydantic? we've used pydantic throughout the app, not sure why dataclass here.

deet/data_models/project.py
nikloynes, line 1: this seems a little in-progress - missing docstrings, and a few design decisions i don't quite understand: env naming/handling, dataclasses

deet/data_models/ui_schema.py
```diff
@@ -0,0 +1,21 @@
+"""
+Definition of data model to add annotations to pydantic model that can be used
+in automatic wizard generation.
+"""
+
+from dataclasses import dataclass
+
+
+@dataclass(frozen=True)
+class UI:
```
nikloynes, line 10: is there a reason why this has to be a dataclass and not pydantic? note that dataclasses are not type safe

deet/extractors/cli_helpers.py
```diff
@@ -4,32 +4,56 @@
...
-        echo_and_log(
-            f"Config file: {config_path} does not exist."
-            " Initialising config with default settings."
+from deet.ui import fail_with_message, notify
+from deet.ui.terminal import console, render_template
+from deet.ui.terminal.components import info_panel
+from deet.ui.terminal.wizards import continue_after_key, run_model_wizard
+
+
+def load_config_from_context(
```
nikloynes: we use the term 'context' several times throughout the app. here a more specific term will be useful.

deet/extractors/cli_helpers.py
```diff
@@ -4,32 +4,56 @@
...
+    ctx: typer.Context, config_path: Path | None
```
nikloynes: no need for abbreviation

deet/extractors/cli_helpers.py
```diff
@@ -4,32 +4,56 @@
...
+        fail_with_message(f"YAML Syntax Error in {config_path}:\n{e}")
+    except ValidationError as e:
+        fail_with_message(f"Config validation error in {config_path}:\n{e}")
 
 
 def init_extraction_run(
     out_dir: Path,
     run_name: str,
-) -> tuple[str, Path]:
+) -> ExperimentArtefacts:
```
nikloynes, line 56: still unclear what this means, so perhaps we can workshop this name.

deet/extractors/cli_helpers.py
```diff
@@ -41,14 +65,14 @@ def init_extraction_run(
 
     logger.add(experiment_out_dir / "deet.log", level="DEBUG")
 
-    return extraction_run_id, experiment_out_dir
+    return ExperimentArtefacts(base_dir=experiment_out_dir, run_id=extraction_run_id)
```
nikloynes, line 68: could do a uuid created from the timestamp instead?

deet/extractors/cli_helpers.py
```diff
@@ -41,14 +65,14 @@ def init_extraction_run(
...
 
 
 def prepare_documents(
     documents: Sequence[Document],
     config: DataExtractionConfig,
     linked_document_path: Path,
-    pdf_dir: Path,
+    pdf_dir: Path | None,
```
nikloynes, line 75: what happens if this is None?

deet/extractors/cli_helpers.py
```diff
@@ -64,7 +88,7 @@ def prepare_documents(
         return documents
     if config.default_context_type == ContextType.FULL_DOCUMENT:
```
nikloynes, line 89: see comment above about context. is this the same 'context' as above, or a different one?

deet/extractors/llm_data_extractor.py
```diff
@@ -100,22 +102,44 @@ class DataExtractionConfig(BaseModel):
...
+    )
+    max_tokens: int | None = Field(
+        default=None,
+        description=(
+            "Maximum number of tokens to generate (None means provider default)."
+        ),
     )
+
     max_context_tokens: int | None = Field(
-        default_factory=lambda: settings.llm_max_context_tokens,
```
nikloynes, line 113: i thought this was a good default. why remove?

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+    def export_config_template(self) -> None:
+        """Export a default config template."""
+        config = DataExtractionConfig()
+        self.config_path.write_text(
+            yaml.safe_dump(config.model_dump(mode="json"), sort_keys=False),
+            encoding="utf-8",
+        )
+
+    @classmethod
+    def load(cls, filename: Path = PROJECT_FILE) -> "DeetProject":
```
VJonah: What would happen if the load method was called outside of a directory with a `project.toml` with no `filename` passed? Wouldn't this cause cascading failures with `PROJECT_FILE` pointing to a non-existent file?

deet/extractors/llm_data_extractor.py
```diff
@@ -100,22 +102,44 @@ class DataExtractionConfig(BaseModel):
...
-            "context). None = infer from model. Override to manage costs."
+            "Maximum input context length in tokens (system + attributes + "
+            "document). None = infer from model (litellm registry), else "
+            f"{DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK} via "
+            "DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK. Override to manage costs."
         ),
     )
+
+    # Context
+    default_context_type: Annotated[
```
nikloynes, line 150: see above

deet/scripts/commands/project.py
```diff
@@ -0,0 +1,139 @@
...
+    render_template,
+    run_model_wizard,
+)
+from deet.ui.terminal.components import info_panel
+
+app = typer.Typer(help="Project-related commands")
+
+
+@app.command()
+def init(ctx: typer.Context) -> None:
```
nikloynes: no need for abbrv IMO ;)

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+PROJECT_FILE = Path("project.toml")
+
+
+class EnvironmentFile(StrEnum):
+    """A choice of where to store settings."""
+
+    PROJECT = auto()
+    SYSTEM = auto()
+
+    def env_file(self) -> Path:
```
VJonah: I don't think this is ever called? And do we need this if we give the user the ability to select system vs local in the wizard that defaults to System?

deet/scripts/commands/project.py
```diff
@@ -0,0 +1,139 @@
...
+    Link documents to their fulltexts.
+
+    This creates a document containing the parsed output of its corresponding
+    fulltext in the folder defined in `output_path`. Linking will be
+    attempted using a mapping file, if provided, then by matching the
+    filename with author and year, then by matching by document id. See
+    `deet.processors.linker` for more details.
+
+    """
+    from deet.processors.linker import DocumentReferenceLinker, LinkingStrategy
```
nikloynes, line 194: have we added ruff exceptions for these kinds of imports for all files? think we should probably add explicitly at the top of files like this where we need.

deet/scripts/commands/project.py
```diff
@@ -0,0 +1,139 @@
...
+    settings = run_model_wizard(DataExtractionSettings)
+    settings.dump_to_env(project.env_path)
+
+    console.clear()
+    console.print(info_panel(render_template("project/success.md", project=project)))
+
+
+@app.command()
+@project_required
+def link(ctx: typer.Context) -> None:
```
nikloynes: more confusion regarding contexts.

deet/scripts/commands/project.py
```diff
@@ -0,0 +1,139 @@
...
+    from deet.processors.linker import DocumentReferenceLinker, LinkingStrategy
+
+    deet_project: DeetProject = ctx.obj.project
+    processed_annotation_data = deet_project.process_data()
+
+    linker = DocumentReferenceLinker(
+        references=processed_annotation_data.documents,
+        document_base_dir=deet_project.pdf_dir,
+        document_reference_mapping=deet_project.link_map_path,
+        linking_strategies=[LinkingStrategy.MAPPING_FILE],
```
nikloynes, line 203: still massively disagree with this. maybe we can have a conversation about how to allow users to auto-link.

deet/scripts/commands/project.py
```diff
@@ -0,0 +1,139 @@
...
+
+from pathlib import Path
+from typing import Annotated
+
+import typer
+from InquirerPy import inquirer
+
+from deet.data_models.project import DeetProject
+from deet.scripts.context import project_required
+from deet.settings import DataExtractionSettings, LogLevel
```
VJonah, line 17: nit: is it worth renaming DataExtractionSettings to DataExtractionSecrets or DataExtractionEnvVars for clarity? Especially if it's only going to contain secrets and seeing as 'settings' and 'config' are synonymous.

deet/ui/terminal/templates/project/success.md
```diff
@@ -0,0 +1,33 @@
...
+
+```sh
+deet run extract
+```
+
+If you prefer to skip the wizard, edit the config file `{{project.config_path}}`
+and pass this as a command line argument
+
+```sh
+deet run --config-path {{project.config_path}} extract
```
VJonah, line 32: nit: shouldn't the `extract` come before the `--config-path` arg?

deet/scripts/commands/project.py
nikloynes: perhaps the auto-linking could do best guesses on how to populate link map for the user, and then they edit it from there.

deet/scripts/typer_context.py
```diff
@@ -0,0 +1,43 @@
...
+
+        ctx = cast(typer.Context | None, raw_ctx)
+
+        if not ctx or not ctx.obj or not ctx.obj.project:
+            no_project = (
+                "This command must be run from a directory that contains a project"
+                " create one by running `deet project init`"
+            )
+            fail_with_message(no_project)
+            raise typer.Exit(1)
```
VJonah, line 50: nit: does this ever get reached if we call `fail_with_message`?

deet/scripts/commands/project.py
```diff
@@ -0,0 +1,139 @@
...
+    from deet.extractors.llm_data_extractor import LLMDataExtractor
+
+    config = load_config_from_context(ctx, config_path)
+    data_extractor = LLMDataExtractor(config=config)
+    attr = Attribute(
+        output_data_type=AttributeType.BOOL,
+        attribute_id=1234,
+        attribute_label="Test Attribute",
+        prompt="Is the document about climate and health? Return a BOOL",
+    )
```
nikloynes, line 246: more confusion about context.

deet/data_models/project.py
```diff
@@ -0,0 +1,268 @@
...
+        return self.base_dir / "goldstandard_llm_comparison.csv"
+
+    @property
+    def prompts_snapshot(self) -> Path:
+        """Return location of csv capturing prompts used."""
+        return self.base_dir / "prompts_used.csv"
+
+    @property
+    def config_snapshot(self) -> Path:
+        """Return location of csv capturing config used."""
```
VJonah, line 244: nit: shouldn't the docstring say `yaml` is returned?

deet/scripts/commands/experiments.py
```diff
@@ -0,0 +1,134 @@
+# ruff: noqa: PLC0415
+"""CLI sub-commands for running data extraction experiments (and evaluating them)."""
+
+from pathlib import Path
+from typing import TYPE_CHECKING, Annotated
+
+if TYPE_CHECKING:
+    from deet.data_models.project import DeetProject
```
nikloynes, line 8: dont think this is really required in >3.13

deet/scripts/commands/experiments.py
nikloynes, line 1: the filename is very non-informative, can do with fleshing out

deet/ui/terminal/wizards.py
```diff
@@ -0,0 +1,166 @@
...
+            prompt_func=prompt_wrapper,
+        )
+        answers[f_name] = answer
+
+    return model_class.model_validate(answers)
+
+
+def continue_after_key(message: str = "Press Enter to continue...") -> None:
+    """Pause execution until the user acknowledges."""
+    inquirer.secret(
```
VJonah, line 211: nit: do we need to use the `secret` method for Enter key presses?

deet/scripts/cli.py
```diff
@@ -1,413 +1,53 @@
...
-    "Prefix any command with --verbose to see complete log output."
-    "will give you more information about how to use the extract-data command.\n\n"
-    "Run `deet --install-completion` to enable your shell to autocomplete deet "
-    "commands."
-)
+app = typer.Typer(help=APP_HELP, add_completion=True, rich_markup_mode="rich")
 
-app = typer.Typer(help=APP_HELP, add_completion=True)
+app.add_typer(project.app, name="project")
+app.add_typer(run.app, name="run")
```
nikloynes: still unclear what this means

deet/scripts/cli.py
nikloynes, line 1: i think the existing commands should be preserved. a lot of users will find them preferable (myself included)

deet/scripts/typer_context.py
```diff
@@ -0,0 +1,43 @@
+"""decorators to handle typer context in commands."""
```
nikloynes, line 1: another one? :o ! see all my other comments... let's please come up with a unique name

deet/scripts/typer_context.py
```diff
@@ -0,0 +1,43 @@
...
+from functools import wraps
+from typing import ParamSpec, TypeVar, cast
+
+import typer
+
+from deet.data_models.project import DeetProject
+from deet.ui import fail_with_message
+
+
+@dataclass
```
nikloynes, line 19: previous comments... why not pydantic?

deet/scripts/typer_context.py
```diff
@@ -0,0 +1,43 @@
...
+    """Structured data store for typer.ctx."""
+
+    project: DeetProject | None = None
+
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def project_required(f: Callable[P, R]) -> Callable[P, R]:
```
nikloynes, line 30: i find this quite difficult to understand, can we maybe rename and add some docstring/comment on what's happening/how it works?

deet/ui/terminal/templates/extraction/config_init.md
nikloynes, line 1: so we're now writing some of our UI instructions/messages in md files? interesting.

deet/ui/terminal/templates/project/configure_env.md
```diff
@@ -0,0 +1,4 @@
+## Let's store your API keys
+
+We will now collect your credentials and save them to a .env file.
```
nikloynes: see previous note about ENV vs .env

also, probably best not to message/echo anything with api keys but rather have the user perform an action, but not clear to me what exactly is happening here.

deet/ui/terminal/templates/help_text.py
```diff
@@ -0,0 +1,22 @@
+"""Help text for CLI commands."""
+
+from deet.ui.terminal.render import flow
+
+APP_HELP = flow("""
+    deet (data extraction evaluation toolkit) 🚤
```
nikloynes, line 6: why the boat emoji?

deet/ui/terminal/templates/help_text.py
nikloynes, line 1: could this also be stuck in an MD file?

deet/ui/terminal/render.py
```diff
@@ -0,0 +1,113 @@
...
+    """Render message to terminal using Rich."""
+    style = level.value.lower()
+    panel = Panel(
+        message,
+        title=f"[{style}]{style.upper()}[/]",
+    )
+    if level in (LogLevel.ERROR, LogLevel.CRITICAL):
+        error_console.print(panel, style=style)
+    else:
+        console.print(panel, style=style)
```
nikloynes, line 48: will this also log to our logfile?

VJonah:
> immediate thoughts/ideas before going through the code:
> 
> > Deet projects can be loaded with DeetProject.load(): loading a deet project assumes that commands are run from a directory containing a valid project toml file.
> 
> * would it perhaps be a good idea to have a central registry in a hidden file in the dir where the app lives, which would mean the user wouldn't have to be in the specific directory?
> * i hope we've still kept the existing CLI args?

- I'm not so sure. We would have to centrally track project names and folder locations alongside some new CLI commands (e.g. updating a project folder's location). All for what feels like a relatively minor UX change. I think enforcing the use of DEET within a project folder doesn't place a meaningful UX burden on the user while reducing complexity. 
- I'm not sure. Max and I discussed building the wizards on top of the existing commands. But I think the command structure has changed with commands having the option to be ran with CLI args instead of a wizard. Maybe @mcallaghan can confirm (I'm looking at functions like `extract` and `link` in the PR)?

deet/ui/terminal/wizards.py
```diff
@@ -0,0 +1,166 @@
...
+from rich.panel import Panel
+
+from deet.data_models.ui_schema import UI
+from deet.ui.terminal import console
+
+T = TypeVar("T", bound=BaseModel)
+UNCHANGED_SECRET = "<unchanged>"  # noqa: S105
+
+
+@dataclass(frozen=True)
```
nikloynes: see previous

deet/ui/terminal/wizards.py
```diff
@@ -0,0 +1,166 @@
...
+
+    field_name: str
+    help_text: str
+    current: int
+    total: int
+
+
+def get_ui_metadata(field_info: FieldInfo) -> UI | None:
+    """Get UI metadata from pydantic model field."""
+    return next((item for item in field_info.metadata if isinstance(item, UI)), None)
```
nikloynes: this seems a bit hacky if it is doing what i think -- why not just pass the right field rather than find it?

deet/ui/terminal/wizards.py
```diff
@@ -0,0 +1,166 @@
...
+def wizard_field_help(field: str, help_text: str) -> None:
+    """Print help text to the right."""
+    help_card = Panel(
+        help_text,
+        title=f"About: {field}",
+        border_style="cyan",
+        padding=(1, 2),
+        width=40,
+        expand=False,
+    )
```
nikloynes: this should probably all go in constants

deet/ui/terminal/wizards.py
```diff
@@ -0,0 +1,166 @@
...
+    def pydantic_validate(answer: str) -> bool | str:
+        try:
+            model_class.__pydantic_validator__.validate_assignment(
+                model_class.model_construct(), field_name, answer
+            )
+        except ValidationError:
+            return False
+        else:
+            return True
+
```
nikloynes, line 167: maybe you can explain the use case here, and add some more docstring. seems to me like we're adding a layer between the user and pydantic, and potentially taking something away pydantic does really well? but it's also not clear to me, so more info will be good.

deet/ui/terminal/wizards.py
```diff
@@ -0,0 +1,166 @@
...
+    widget_args: dict[str, Any] = {
+        "message": field_info.description,
+        "default": default,
+        "validate": lambda ans: pydantic_validate(ans),
+        "invalid_message": ui.valid,
+        "instruction": ui.instructions,
+        "filter": lambda ans: ans.strip(),
+    }
+
+    if isinstance(field_info.annotation, type) and issubclass(
```
nikloynes: surely we can refactor this with a strategy pattern, or a map, something like that?

deet/ui/terminal/wizards.py
```diff
@@ -0,0 +1,166 @@
...
+    total_steps = len(ui_steps)
+
+    for i, (f_name, f_info, f_ui) in enumerate(ui_steps, 1):
+
+        def prompt_wrapper(
+            field_name: str = f_name,
+            field_info: FieldInfo = f_info,
+            ui: UI = f_ui,
+        ) -> str | None:
+            return inquire_pydantic_field(model_class, field_name, field_info, ui)
```
nikloynes: not clear what this is doing -- could do with more documentation and info

deet/settings.py
```diff
@@ -25,6 +29,18 @@ class LLMProvider(StrEnum):
     OLLAMA = auto()
 
 
+class LogLevel(StrEnum):
```
nikloynes, line 32: this already exist, i don't think we need to re-define

deet/settings.py
```diff
@@ -41,62 +57,59 @@ class DataExtractionSettings(BaseSettings):
     """
 
     model_config = SettingsConfigDict(
-        env_file=".env", env_file_encoding="utf-8", extra="ignore"
+        env_file=[".env", Path.home() / ".deet" / ".env"],
```
nikloynes: note my previous thoughts about ENV

deet/settings.py
```diff
@@ -41,62 +57,59 @@ class DataExtractionSettings(BaseSettings):
...
-            "document). None = infer from model (litellm registry), else "
-            f"{DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK} via "
-            "DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK. Override to manage costs."
-        ),
+        json_schema_extra={"skip_prompt": True},
     )
 
     # Provider credentials / settings (secrets redacted)
-    azure_api_key: SecretStr | None = Field(
+    azure_api_key: Annotated[SecretStr | None, UI()] = Field(
```
nikloynes: you'll have to explain what's going on here with the UI thing

deet/settings.py
```diff
@@ -41,62 +57,59 @@ class DataExtractionSettings(BaseSettings):
...
 
     # disk cache folder
     base_disk_cache_dir: Path = Field(
         default=(Path.home() / ".deet_cache"),
         description="the base directory for disk-based caches.",
+        json_schema_extra={"skip_prompt": True},
     )
 
+    def dump_to_env(self, target_path: Path) -> None:
+        """Serialise settings object to a .env file."""
```
nikloynes, line 97: if this is for project dir stuff, i think we'd be best not to call it .env, see previous notes.

tests/unit/test_ui/test_messenger.py
nikloynes, line 1: i'll review tests later on

nikloynes (commented):
huge PR. think this should really be the limit of what a PR encompasses, especially as it covers quite a few different concerns. 

having said that, this looks really promising. i have a few comments on general design implementation, but they're minor. i think the next iteration can iron out some small stuff, especially naming of objects & docstrings/documentation, and thinking about whether yet another import/different format is always required (toml, dataclasses), or can just use a flow we've used previously in the application.

i think overall this is now making the application very opinionated in its conceptualisation and design. it seems that parts of the research team are in favour of this, and i am definitely in favour of ease of use. i do however think, some of these choices might make contributions and modifications further down the road harder, and maybe the next iteration of this PR can also just keep those ideas of flexibility and maintainability for the future in mind as guiding principles. i'm thinking specifically of following features/use-cases: 
- data extraction w/o gold standard data
- prompt versioning

one thing i'd say is missing, is a brief explainer of how to use the wizard (i don't love this term btw, seems very windows XP -- but i get that it's the right one). maybe even a little gif/video. 

note that i haven't reviewed tests as of yet, will keep that for next round.

mcallaghan:
>     * data extraction w/o gold standard data

In my eyes, this is still as possible as before. Your eppijson file (or other) which you use to initiate your project would simply define the attributes and the documents, but, in eppijson terms, "Codes" for each document would simply be null (I think we still need to better distinguish between empty and bool=False annotations, but that's another story). Once we have other importers, we could think of having an "attribute-only" import format, and having a command to create `Document` objects from pdfs directly. I think this is what you are trying to make sure is still possible? But again I think these workflows are no more or less possible with the changes here. At least I'm definitely not trying to exclude these.



> * prompt versioning

Yes I think this is actually easier than before, in that the prompts used for each experiment/run are stored in the directory created for the run. `deet.data_models.project.ExperimentArtefacts` describes what is put in that directory.

This means we can easily compare the prompts used between experiments (and the metrics achieved). This makes it easy to think of a command along the lines of "look through my experiments and show me the prompt for attribute X that achieved the highest recall score", or a command that says "restore prompts to those used in experiment Z", or "do a new experiment with a different model, but the same prompts as were used in experiment 5".

I had first thought of letting the user create multiple prompt definition files, and have the config / CLI args point to one of these. Then when the config is dumped in the experiment artefacts directory, you could read from this that this run was created with `prompts/prompt_definitions_3_final_FINAL (1).csv`, but I think there is a real risk that users accidentally overwrite data.

Instead, each project has a single user-editable prompt definition file, and every time this is used for data extraction, we take a snapshot of the prompts used and dump in the experiment directory.

I thought about doing something fancy with hashes, but think this can be left for future PRs :)

mcallaghan:
> immediate thoughts/ideas before going through the code:
> 
> > Deet projects can be loaded with DeetProject.load(): loading a deet project assumes that commands are run from a directory containing a valid project toml file.
> 
>     * would it perhaps be a good idea to have a central registry in a hidden file in the dir where the app lives, which would mean the user wouldn't have to be in the specific directory?

I agree with Jonah, I think navigating to the directory where you are working is a common enough pattern, but perhaps we could think of that for the future?

>     * i hope we've still kept the existing CLI args?

To a large extent yes! Some are now made redundant by the fact you are in a project. Once you have created a project, and run commands from that project, you no longer have to specify the gold standard data, the gold standard data format, the pdf directory, the prompt file (see above), and the link map. For the main extract command, you can skip the wizard by providing a config file instead, so it's still possible to run in non-interactive mode (but that reminds me, there are a couple of press enter to continue instances that should either timeout or be automatically skipped if not in interactive mode)

deet/data_models/project.py
mcallaghan: Not sure I fully understand this, but a pragmatic solution would be to revert to a .env file in each project, created on project init. 

What I wanted to achieve with this was to allow people to easily re-use credentials across deet projects, but also to be able to specify specific keys for specific projects. But maybe we create an issue for credential re-use and keep it simple here?

deet/data_models/project.py
mcallaghan: Why not just require this? Seems like a meaningful name is a useful requirement. The wizard will ask for a name, and will apply the pydantic validators, so won't accept empty or one letter names.


Another option would be to take the name of the parent directory as the name, or to have name as an optional argument of the command. E.g., from `~/projects/p1` `deet project init` creates a project in p1 called p1, and `deet project init new-proj` creates a project new-proj in `~/projects/p1/new-proj`, and cds into this directory? A bit like the behaviour of uv init?

deet/data_models/project.py
mcallaghan: This is just the default, so a user can certainly override this. As it does not have UI metadata, this won't be part of the wizard, so it won't be easy to override this in the CLI, but I don't think it needs to be. (User could always just edit the project file)

deet/data_models/project.py
mcallaghan: what are alternatives? This is encoding that if you create a project from `~/projects/p1`, then the project root will be in `~/projects/p1`. All project artefacts (prompt definition file, pdf link map, etc.) will live in that directory.

deet/data_models/project.py
mcallaghan: for me `project.setup()` seems clear enough, `project.setup_project()` tautological, but happy to defer

deet/data_models/project.py
mcallaghan: yes this is probably overkill. toml struck me as more project-definitiony and yaml as more configurationy, but I think we can stick to one of these (probably yaml) for anything that a human might want to look at or edit, and json for everything that's just for machines. What do you reckon?

deet/data_models/project.py
mcallaghan: so we should use pydantic for everything? My understanding was only if we have messy data that requires validation. Happy to convert this and the other dataclasses to pydantic, but I wondered, does it actually need to be a pydantic model here

deet/data_models/project.py
nikloynes: .env is a tool for development, to mimic environment variables without having to set them on your machine. i think copying the file is fine, but keeping it as .env stretches the concept a bit too far

deet/data_models/project.py
nikloynes: because, my theory is, ppl will call it 'my-test-folder-17' and get confused. if it's called 'sherbet-monkey' instead, they'll find it easier to remember.

deet/data_models/project.py
nikloynes: sorrry, i missed that this was in a class, thought it was a function.

deet/data_models/project.py
nikloynes: i think they are used for very similar things and introducing dataclasses here seems, to me, somewhat confusing to the casual contributor and a bit stylistically inconsistent. i don't think all our pydantic models necesary _need_ to be pydantic... i just _personally_ don't like mixing this. if it's important to you, keep it in there.

deet/data_models/ui_schema.py
mcallaghan: see above, this is not handling any external data, and since we are using mypy, if we ever tried to annotate a pydantic model with UI(help=123), mypy would catch this. Don't think we need to enforce type checking at runtime. 

There will be an overhead to using pydantic, but if you think this is worth it to simplify our lives by having a rule "use pydantic for everything" then I'm happy to go with that

deet/extractors/cli_helpers.py
mcallaghan: yes good point. This type of context is typer context: https://typer.tiangolo.com/tutorial/commands/context/#getting-the-context

think we can mostly replace `context` with `typer_context` whenever we are talking about this type of context

deet/extractors/cli_helpers.py
mcallaghan: I'm following the conventions here: https://typer.tiangolo.com/tutorial/commands/context/, but maybe this is clearer if we always call this `typer_context` instead (although this is clear enough from the type hint)?

deet/extractors/cli_helpers.py
mcallaghan: I thought we prefered the human-readable name?

deet/extractors/cli_helpers.py
mcallaghan: if `config.default_context_type` (not the typer context, but the kind defined in `deet.data_models.documents.ContextType` == "abstract_only", then we continue as normal, no need to link anything, we just return the documents as they are, otherwise DocumentReferenceLinker will raise an error. We should probably catch this one step previously and exit gracefully with a clear message

deet/extractors/cli_helpers.py
mcallaghan: see above, not the typer one

deet/scripts/commands/project.py
mcallaghan: I think we've got too many of these types of imports and what would probably make sense would be to identify which externally imported libraries are the ones causing big slowdowns, and to import these lazily, then go back to as a rule importing at the top of files

deet/scripts/commands/project.py
mcallaghan: Yes I think something along these lines could work nicely, and we could even do this via a wizard, which would have the benefit of being able to validate that the file paths entered by a user exist

deet/scripts/commands/experiments.py
mcallaghan: ruff told me to do it! And I think we're in 3.12

deet/scripts/commands/experiments.py
mcallaghan: I think we should have a conversation about the names and structure of commands, but we can also have them terse in the cli and longer and descriptive in the filenames

deet/scripts/cli.py
mcallaghan: Are you happy to move these into subcommands? At the moment most commands are moved, but two commands (init link map, and init prompt csv) are subsumed into project setup. I think we need to find a way to make things less overwhelming by moving things into subcommands, and I think we should take the opportunity now before we have a properly stable version

deet/scripts/typer_context.py
mcallaghan: it's a decorator for cli commands that checks for the presence of a project and exits if not. I'll update the docstring

deet/ui/terminal/templates/extraction/config_init.md
mcallaghan: yes, pleased with this :)

deet/ui/terminal/templates/project/configure_env.md
mcallaghan: It doesn't echo API keys! This is just an introduction to the wizard, which will use a [secret prompt](https://inquirerpy.readthedocs.io/en/latest/pages/prompts/secret.html)

deet/ui/terminal/templates/help_text.py
mcallaghan: unfortunately I couldn't figure out how to do this, but agree this would be preferable. Elsewhere we use rich.markdown to render markdown to the terminal, but though the help text is rendered with rich, I can't find a way for it to render markdown

deet/ui/terminal/templates/help_text.py
mcallaghan: because deet is a "speedboat", but happy to change

deet/ui/terminal/render.py
mcallaghan: no! This bit just handles the terminal `deet.ui.messenger.notify` sends messages to logger, console, and in the future anywhere else we want to send it to (for toasts in a web ui)

deet/ui/terminal/wizards.py
mcallaghan: It's getting the UI metadata object from the field, which is stored as a list. So for a field annotated like so:

```python
name = Annotated[
    str, UI(help="useful help text to display in my wizard")
]
```
it will return the UI object, which we can then use to make the wizard nice

deet/ui/terminal/wizards.py
mcallaghan: or rather in `deet.ui.terminal.components` ?

deet/ui/terminal/wizards.py
mcallaghan: it uses the validator defined in the pydantic object to prevent the user from being able to return input that is not valid, so if they type `ab` for a field with `min_length=3`, instead of inquirer accepting this, and then pydantic raises a validation error, the user will see a message saying "not long enough", and they won't be able to move on until they enter valid data of exit

deet/ui/terminal/wizards.py
mcallaghan: yes was not 100% trivial as some field types need some additional arguments but can probably come up with something

deet/ui/terminal/wizards.py
mcallaghan: Think I might be able to simplify and clarify this file, will have a go tomorrow

deet/settings.py
mcallaghan: where is this already defined? Previously we used a string. and making this an enum makes this type safe

deet/settings.py
mcallaghan: If UI metadata is present, then it is included in a wizard, otherwise the wizard skips the field when creating prompts. This should have some help text though!

mcallaghan:
Just finished going through your individual comments, thanks very much! Think there are a few meta-points:

## environment variable management

This obviously needs more thinking through, I suggest we revert to collecting settings via a wizard, and then put them in a `.env` file in the project, and leave anything system-wide for another PR

## command naming and organisation

I think we should discuss this, and come to an agreement on subcommands and names and get this close to stable. For me I think having shorter names, and grouping things into subcommands is a UX win, especially for those not used to the terminal using tab to autocomplete

## documentation

I've added and expanded docstrings here and there, but I think this would benefit from adding to the actual docs - think it's time to add different sections to this for different audiences at different levels, one the one hand for cli users (including with a gif of the wizard), and on the other at a level between this and the auto-generated API documentation that explains the modules and concepts in more detail.

deet/ui/terminal/wizards.py
nikloynes: sure

deet/settings.py
nikloynes: either in the logging module - logging.log_level or something like that, or in loguru, which wraps around logging and we use here

mcallaghan:
Pending further documentation, here's a gif of the wizard in action. Have just figured out how to generate these programmatically for the docs with vhs 

project_init

VJonah (changes requested):
A lot of great stuff here, particularly with the wizards and documentation. A few nits, nothing major. But I think I agree with Nik that it might be worth splitting the PR into separate issues (e.g. UI/wizard improvements, and config vs setting refactors, etc) for clarity, simplicity, and in the spirit of atomic changes.

deet/data_models/project.py
VJonah: Unless we're centrally managing project names and their directories, when will a user need to remember the name of their project? If we have random project names why have them in the first place?

deet/data_models/project.py
VJonah: Is there a particular reason a user would want to change the project creation date?

deet/data_models/project.py
VJonah: I agree with Nik. I don't think we lose anything from pydantic, there's only upside and we can mute validation in places we see fit.

deet/data_models/ui_schema.py
VJonah: What's the overhead to using pydantic sorry? Performance?

deet/settings.py
mcallaghan: Apparently the logging module predates enums: [random blog link](https://hrekov.com/blog/python-logging-levels-enum), [stack overflow](https://stackoverflow.com/questions/75302200/python-types-literal-of-logging-level-as-type) (can't find much else by googling "python logging enum"), so I think if we want this as an enum (which is an improvement over typing this as a string, which wouldn't catch `inf0` for example), then we need to write the enum ourselves, but convention seems to be rather to do it like this: `INFO = logging.INFO`

deet/data_models/ui_schema.py
mcallaghan: yes, see for example: https://leehanchung.github.io/blogs/2025/07/03/pydantic-is-all-you-need-for-performance-spaghetti/. I don't think things need to be super-fast but I'm feeling the slow startup for `deet --help` for example.

deet/extractors/llm_data_extractor.py
mcallaghan: `settings.llm_max_context_tokens` used to also default to None. Now we simply define this default=None field in one place and one place only (DataExtractionConfig)

deet/scripts/commands/project.py
mcallaghan: But I'll mark as resolved as this is not part of this PR

deet/data_models/project.py
mcallaghan: I'll close this and we leave as it is for now, can return to this later

deet/data_models/project.py
mcallaghan: #219

deet/data_models/project.py
mcallaghan: I don't think so, if they really wanted to, they could edit the project yaml file, but I don't think this would be wise. Resolving..

deet/settings.py
mcallaghan: since loguru has it's own set of levels, I think the way this is in here is actually the best way to do it.

deet/settings.py
mcallaghan: this is just to save a settings object to a .env file, so we can create a settings object by prompting for keys, and then write this to .env, to save users having to edit this file themselves (and I think dealing with hidden files can be intimidating for the casual user)

mcallaghan: This now also only goes to .env in the current directory, so closing

deet/data_models/project.py
mcallaghan: clarified in docstring

deet/scripts/commands/experiments.py
VJonah: I get it's lazy loading for performance now, sorry

deet/extractors/cli_helpers.py
mcallaghan: This is now properly handled and tested

deet/ui/terminal/wizards.py
mcallaghan: this is now done

deet/ui/terminal/wizards.py
mcallaghan: Should hopefully be clearer now, definitely benefited from a rewrite with fresh eyes

VJonah (approved):
I think this is self-contained enough and doesn't introduce any breaking or backend changes. Happy to merge.

PR #229, opened 2026-04-15
Feature/evaluator fails gracefully

### LINKED ISSUE #228: Fail more gracefully when trying to evaluate
### Description

The evaluator raises an error when trying to a retrieve a human or llm annotation for a document and, either

- The document isn't in the list of annotated documents (perhaps because linking failed and no llm annotation was produced)  
- There are multiple annotations for the given attribute, and we don't know how to resolve which one is correct.

When this error is raised, the command exits, and none of the experiment results are saved.

### Definition of Done

- [ ] When encountering one of the above errors the `extract-data` continues to run to the end
- [ ] A warning is shown to the logger
- [ ] The metric trying to be calculated receives a None/NA
- [ ] The comparison sheet gets an empty cell and an appropriate explanation in the reasoning column

### Dependencies

- #Issue

mcallaghan:
# Pull Request

## Description
In evaluation code, looking up LLM annotations resulted in an error if an LLM annotation was not produced, or if the LLM produced more than one annotation for the same attribute.

Now, this causes a warning, and for metrics to evaluate to None, but the command finishes running and outputs are generated

## Related Issue(s)
Closes #228 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Testing
Wrote test fixtures for mising/duplicated llm annotation data, and tests for how we want it to behave. 

## Additional Notes

---

### DIFF
--- a/deet/evaluators/gold_standard_llm_evaluator.py
+++ b/deet/evaluators/gold_standard_llm_evaluator.py
@@ -7,6 +7,7 @@ import csv
 from collections.abc import Sequence
 from itertools import groupby
 from pathlib import Path
+from typing import Any
 
 import sklearn.metrics  # type:ignore[import-untyped]
 from loguru import logger
@@ -94,7 +95,7 @@ class GoldStandardLLMEvaluator:
                 f"Calculating metric for attribute: {attribute.attribute_label}"
             )
             y_true = []
-            y_pred = []
+            y_pred: list[Any] = []
             for document in self.gold_standard_annotated_documents:
                 logger.debug(
                     f"Extracting gold standard and LLM prediction for"
@@ -103,9 +104,15 @@ class GoldStandardLLMEvaluator:
                 gs_val = document.get_attribute_annotation(attribute).output_data
                 y_true.append(gs_val)
 
-                llm_doc = self.llm_annotated_documents.get_by_id(
-                    document.document.safe_identity.document_id
-                )
+                try:
+                    llm_doc = self.llm_annotated_documents.get_by_id(
+                        document.document.safe_identity.document_id
+                    )
+                except ValueError:
+                    y_pred.append(None)
+                    missing_id = document.document.safe_identity.document_id
+                    logger.warning(f"LLM annotated doc not found - ID: {missing_id}")
+                    continue
                 llm_val = llm_doc.get_attribute_annotation(attribute).output_data
                 y_pred.append(llm_val)
 
@@ -117,6 +124,7 @@ class GoldStandardLLMEvaluator:
             for metric_name, metric_fn in combined_metrics.items():
                 try:
                     value = float(metric_fn(y_true, y_pred))
+                    logger.warning(f"It worked for {metric_name}!")
                 except (ValueError, TypeError) as e:
                     logger.warning(
                         f"Metric '{metric_name}' not applicable for "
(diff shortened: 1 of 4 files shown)


deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -196,10 +208,32 @@ def export_llm_comparison(
...
+                        try:
+                            llm_annotation = llm_annotated_doc.get_attribute_annotation(
+                                attribute
+                            )
+                            llm_extraction = llm_annotation.output_data
+                            llm_reasoning = llm_annotation.reasoning
+                        except ValueError:
+                            llm_extraction = None
+                            llm_reasoning = (
+                                "The LLM produced multiple annotations"
```
L-ENA, line 234: I think the previous code handled the value error directly in documents.py in the get_attribute_annotation, while this edit makes sure that DEET then continues and adds information to the output for the user. The only question that came to my head is: Can we be sure that multiple matches are the only reason for a Value error at this stage, or do we risk telling the user that something is a value error for multiple matches while there may be other reasons that the prediction can fail? Maybe this is something to consider after merging the PR, because broad usage across many datasets will show us how and where things break

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -117,6 +128,7 @@ def evaluate_llm_annotations(
             for metric_name, metric_fn in combined_metrics.items():
                 try:
                     value = float(metric_fn(y_true, y_pred))
+                    logger.warning(f"It worked for {metric_name}!")
```
L-ENA: This caused an extremely high amount of logger WARNINGs on my small test dataset, which are essentially success messages that the metrics were computed. While this is useful, it also means that the more rare events (such as the warning where the metric wasn't computed, might be drowned out). Might it be easier for the user to spot issues in the log if it alerts only to the failures? The log already tells us that it is computing each metric so it could be implicit that the computation is a success if there isn't an error warning for the metric.

L-ENA (approved):
Thank you for implementing this over night, @mcallaghan .
I like how the value error is caught, how None is added instead, and how the new code intruduces an 'artificial' LLM reasoning string to alert the user about the problem and to point them to the log. This ensures transparancy about the LLM's failure to process the request, and is good to debug things for extreme cases where users don't check their predictions. 

I was able to reproduce the Value error for multiple matches. The log and the comparison CSV in your fix both pointed out the issue, which means that the fix worked :)

I tested how the new code handles the error where no LLM annotations are produced due to a too small context window (while truncate_on_overflow: false). This worked very well and the comparison csv as well as useful outputs were still produced for those records that did not exceed the token amounts. 

One small comment in the code, but this can also be deferred and dealt with once we have a better idea of the varieties of errors caused by the LLM or our downstream datasets.

Something I noticed when looking at the log was that the evaluator creates a lot of messages. I added a small comment in the code, but these might also be dealt with later or after discussion with the team. There is another issue I found with the log which isn't related to this PR, in deet.data_models.evaluation:save_to_csv: My log ends with 1254 times exactly the same message, which might be excessive. It is the same message only with slightly different timestamp: "2026-04-15 08:49:36.985 | DEBUG    | deet.data_models.evaluation:save_to_csv:132 - Wrote metric to data-extraction-experiments\2026-04-15_07-42-27_\metrics.csv" We can discuss this in one of the meetings if needed because it does not directly relate to this PR

Feel free to merge I guess?

deet/evaluators/gold_standard_llm_evaluator.py
mcallaghan: This is a good point, I've added custom exceptions to make sure we only catch missing documents or duplicate annotations

deet/evaluators/gold_standard_llm_evaluator.py
mcallaghan: oops, that was just there to help me debug my tests! We definitely don't need this

PR #239, opened 2026-04-22
Fix/234 change annotation data type

### LINKED ISSUE #234: Investigate how we store annotations after user specifies type in prompt csv
### Description
When we specify the type of attributes in our prompt csv, `deet.data_models.processed_gold_standard_annotations.ProcessedAttributeData._process_csv_row()` updates the data type of the relevant attribute, but does not update the annotations. 

This can leave us in a situation where we have an annotation with output_data=True, and attribute.annotation_type = "string"

### Definition of Done

- [ ] Changing the type of an attribute in the prompt csv changes the output_data field of annotations with that attribute

mcallaghan:
# Pull Request

## Description

When an attribute's type is changed, e.g. by processing a custom prompt csv, output_data reflects the change for each annotation that refers to that attribute.

This is done by
- replacing the `output_data` field with a `raw_data` field in `deet.data_models.base.GoldStandardAnnotation`. When an annotation is created, we enter the value from the source data into raw_data, without any processing based on the assumed data type. This always stays the same.
- creating `output_data` as a computed field. This coerces the raw_data into the format determined by attribute's type.

This test (from `tests/unit/test_base.py`) demonstrates how this behaviour works:

```python
def test_gold_standard_annotation_string_preserved_on_type_change() -> None:
    """Test that wrong type for bool attribute raises ValueError."""
    attr = Attribute(
        prompt="Is this valid?",
        output_data_type=AttributeType.BOOL,
        attribute_id=1234,
        attribute_label="Test Bool Attribute",
    )

    string_value = "not a bool"
    annotation = GoldStandardAnnotation(
        attribute=attr,
        raw_data=string_value,
        annotation_type=AnnotationType.HUMAN,
    )
    # The following is true because even though we gave the raw_data value "not a bool", 
    # we said it had ouput_data_type bool, so output_data is coerced to a boolean True
    assert isinstance(annotation.output_data, bool) 
    
    # If we change the attribute type of the annotation's attribute
    attr.output_data_type = AttributeType.STRING

    # Then it is no longer a bool but a string, with it's original value that was stored in raw_data
    assert isinstance(annotation.output_data, str)
    assert annotation.output_data == string_value
```

Coercion strategies are catalogued in `deet.data_models.base.ANNOTATION_COERCION_STRATEGIES`. At the moment, we only have strategies for bool, string, int and float. We can think of strategies for other types. E.g. a list strategy can split the string with a few different separator patterns. But we can leave these for further work, as these will require thinking.

For backwards compatibility, if we try to create an annotation with `output_data="abc"`, this will be redirected to `raw_data`

## Related Issue(s)
Closes #234 

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Additional Notes
Note that this change affects both human and llm annotations. Because the presence of an annotation for a boolean eppi attribute indicates Truth, any text that is not falsey ("0", "false") will be coerced to True.

---

### DIFF
--- a/deet/data_models/base.py
+++ b/deet/data_models/base.py
@@ -267,7 +267,7 @@ def coerce_annotation_to_bool(val: SUPPORTED_TYPES) -> bool:
 
 
 def coerce_annotation_to_int(val: SUPPORTED_TYPES) -> int | None:
-    """Coerce an annotation to a bool."""
+    """Coerce an annotation to a int."""
     if isinstance(val, int):
         return val
     if isinstance(val, str | float | bool):
@@ -276,21 +276,21 @@ def coerce_annotation_to_int(val: SUPPORTED_TYPES) -> int | None:
         except ValueError:
             logger.warning("Could not convert {val} to int")
 
-        logger.warning("Could not convert {val} to int")
+    logger.warning(f"Unsupported type for int conversion: {type(val).__name__}")
     return None
 
 
 def coerce_annotation_to_float(val: SUPPORTED_TYPES) -> float | None:
-    """Coerce an annotation to a bool."""
+    """Coerce an annotation to a float."""
     if isinstance(val, float):
         return val
     if isinstance(val, str | bool | int):
         try:
             return float(val)
         except ValueError:
-            logger.warning("Could not convert {val} to float")
+            logger.warning(f"Could not convert {val} to float")
 
-        logger.warning("Could not convert {val} to float")
+    logger.warning(f"Unsupported type for float conversion: {type(val).__name__}")
     return None
 
 
--- a/deet/data_models/base.py
+++ b/deet/data_models/base.py
@@ -254,8 +254,14 @@ def coerce_annotation_to_str(val: SUPPORTED_TYPES) -> str:
     return str(val) if val else ""
 
 
-def coerce_annotation_to_bool(_val: SUPPORTED_TYPES) -> bool:
+def coerce_annotation_to_bool(val: SUPPORTED_TYPES) -> bool:
     """Coerce an annotation to a bool."""
+    if isinstance(val, str) and val.lower() in ("false", "0"):
+        return False
+
+    if isinstance(val, int | float):
+        return bool(val)
+
     return True
 
 
--- a/deet/data_models/base.py
+++ b/deet/data_models/base.py
@@ -67,8 +67,7 @@ class AttributeType(StrEnum):
                 return {}
             case _:
                 unsupported = (
-                    "No default for missing annotation when attribute type is "
-                    f"{self!s}"
+                    f"No default for missing annotation when attribute type is {self!s}"
                 )
                 raise ValueError(unsupported)
 
@@ -368,7 +367,7 @@ class LLMAnnotationResponse(BaseModel):
         ...,
         description="The LLM's annotation.",
         json_schema_extra=cast(
-            JsonDict,
+            "JsonDict",
             {
                 "anyOf": [
                     attribute_type.to_json_type() for attribute_type in AttributeType
--- a/deet/data_models/base.py
+++ b/deet/data_models/base.py
@@ -303,7 +303,13 @@ ANNOTATION_COERCION_STRATEGIES: dict[AttributeType, Callable] = {
 
 
 class GoldStandardAnnotation(BaseModel):
-    """A single gold standard annotation for an attribute."""
+    """
+    A single gold standard annotation for an attribute.
+
+    `raw_data` stores the data as it comes from source,
+    `output_data` is computed and coerces raw_data into the correct type.
+    This can change if the `AttributeType` of the `attribute` changes.
+    """
 
     attribute: Attribute
     raw_data: Any = Field(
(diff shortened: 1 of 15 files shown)


deet/data_models/base.py
```diff
@@ -246,11 +248,70 @@ def sanitize_prompt(prompt: str) -> str:
 AttributeTypeVar = TypeVar("AttributeTypeVar", bound=Attribute)
 
 
+def coerce_annotation_to_str(val: SUPPORTED_TYPES) -> str:
```
sagaruprety, line 251: In future will we need to do something for list/dict types as well? Or should add a placeholder now itself?

deet/data_models/base.py
```diff
@@ -246,11 +248,70 @@ def sanitize_prompt(prompt: str) -> str:
...
+    AttributeType.FLOAT: coerce_annotation_to_float,
+}
+
+
 class GoldStandardAnnotation(BaseModel):
     """A single gold standard annotation for an attribute."""
 
     attribute: Attribute
-    output_data: Any
+    raw_data: Any = Field(
```
sagaruprety, line 317: can we have a explicit note in the docstring (and maybe in the PR description) that output_data is now derived and may change if attribute.output_data_type changes, while raw_data is the immutable original.

deet/data_models/base.py
```diff
@@ -246,11 +248,70 @@ def sanitize_prompt(prompt: str) -> str:
...
+    """Coerce an annotation to a bool."""
+    if isinstance(val, int):
+        return val
+    if isinstance(val, str | float | bool):
+        try:
+            return int(val)
+        except ValueError:
+            logger.warning("Could not convert {val} to int")
+
+        logger.warning("Could not convert {val} to int")
```
sagaruprety: duplicate logging?

deet/data_models/base.py
```diff
@@ -246,11 +248,70 @@ def sanitize_prompt(prompt: str) -> str:
...
+    """Coerce an annotation to a bool."""
+    if isinstance(val, float):
+        return val
+    if isinstance(val, str | bool | int):
+        try:
+            return float(val)
+        except ValueError:
+            logger.warning("Could not convert {val} to float")
+
+        logger.warning("Could not convert {val} to float")
```
sagaruprety: duplicate logging

deet/data_models/base.py
```diff
@@ -246,11 +248,70 @@ def sanitize_prompt(prompt: str) -> str:
...
+    if isinstance(val, str) and val.lower() in ("false", "0"):
+        return False
+
+    if isinstance(val, int | float):
+        return bool(val)
+
+    return True
+
+
+def coerce_annotation_to_int(val: SUPPORTED_TYPES) -> int | None:
```
sagaruprety, line 269: this function accepts  str | float | bool and str | bool | int respectively, but the return type allows None. Just wondering if this wont be an issue for any caller or any functionality down the line which assumed a concrete type but would now also see None.

deet/data_models/base.py
```diff
@@ -246,11 +248,70 @@ def sanitize_prompt(prompt: str) -> str:
...
+        try:
+            return float(val)
+        except ValueError:
+            logger.warning("Could not convert {val} to float")
+
+        logger.warning("Could not convert {val} to float")
+    return None
+
+
+ANNOTATION_COERCION_STRATEGIES: dict[AttributeType, Callable] = {
```
sagaruprety: can we do  Callable[[SUPPORTED_TYPES], SUPPORTED_TYPES] so that we retain type info about the input/output?

tests/unit/test_base.py
```diff
@@ -786,12 +786,37 @@ def test_gold_standard_annotation_bool_type_invalid() -> None:
...
+        raw_data="not a bool",
+        annotation_type=AnnotationType.HUMAN,
+    )
+
+    assert isinstance(annotation.output_data, bool)
+    assert annotation.output_data
+
+
+def test_gold_standard_annotation_string_preserved_on_type_change() -> None:
+    """Test that wrong type for bool attribute raises ValueError."""
```
sagaruprety: does the docstring need updating?

sagaruprety (changes requested):

deet/data_models/base.py
mcallaghan: I think we shoud leave this for later so we can discuss how to implement. We have an open issue for list and dict, so let's make sure we add a strategy in there. As it is we mirror current behaviour (specifying list doesn't do anything to annotations).

deet/data_models/base.py
mcallaghan: yes! was supposed to be outside the if block, and the message now clarifies the difference (type can in principle be converted, but not this value vs type cannot be converted)

deet/data_models/base.py
mcallaghan: good point, I've changed output_data, specifying that it can also return None (in case of unconvertible data). This should mean downstead uses are warned.

I think None is preferable to raising an exception, as per previous discussions on errors. None allows a run to continue, and metrics just won't work = minimal disruption to user

deet/data_models/base.py
mcallaghan: nice, yes! Or rather SUPPORTED_TYPES | None (see previous comment)

tests/unit/test_base.py
mcallaghan: yes it does :). Done now!

sagaruprety (approved):
I think we can merge it now, thanks for addressing my comments.

PR #240, opened 2026-04-23
introduce dynamic versioning

### LINKED ISSUE #207: Determine process for versioning updates once v1 release public
### Description
Need to ensure that subsequent versions are clearly documented (in plain language) for users once first release is public

### Definition of Done

- [x] Decide sharing mechanism for updates
- [x] Decide on who is responsible for writing the plain language update 
- [ ] Decide a threshold for when an update is required 

### Dependencies

- #Issue

mcallaghan:
# Pull Request

## Summary of versioning strategy:

To create a version, we simply change UNRELEASED at the top of CHANGELOG to the new version number, merge everything into main, and then do `git tag [version number]`. Later on, we can create a github action to deploy to pypi (if that's what we want) when we tag an new version 

## Description

- Use setuptools_scm to derive versions dynamically from git tags.
- Git Tag the last version change, so current version appears as 0.2.0.dev383... indicating the number of commits since the last version
- Create CHANGELOG.md and fill a basic baseline (we should be more detailed going forwards, but I don't think we need to re-create the whole history now)
- Update CONTRIBUTING.md and pull request template to describe how to keep CHANGELOG up-to-date, and describe versioning policy
- Add `deet --version` / `deet -V` to print current version of CLI.

## Related Issue(s)

Closes #207 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Additional Notes

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -224,3 +224,6 @@ default_extraction_config.yaml
 pdfs/*
 data-extraction-experiments/*
 *.csv
+
+# Dynamic version
+deet/_version.py
--- a/.gitignore
+++ b/.gitignore
@@ -224,6 +224,3 @@ default_extraction_config.yaml
 pdfs/*
 data-extraction-experiments/*
 *.csv
-
-# Dynamic version
-deet/_version.py
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -4,3 +4,5 @@ MD013: false #  line-length
 ignore:
   - "tests/test_files/**/*"
 MD046: false # code block style (need both styles if tabs are to work)
+MD024:
+  siblings_only: true
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -4,5 +4,5 @@ MD013: false #  line-length
 ignore:
   - "tests/test_files/**/*"
 MD046: false # code block style (need both styles if tabs are to work)
-MD024:
+MD024: # allow duplicated headings as long as they are not siblings (for changelog)
   siblings_only: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,48 @@
+# Changelog
+
+All notable changes to this project will be documented in this file
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [0.3.0]
+
+### Added
+
+- CLI to perform parsing, processing and extracting data
+- Linking of records in goldstandard annotation data to pdfs
+- Evaluation of data extracted by LLMs
+- Documentation
+- Parsing of annotation data in CSV format
+
+### Changed
+
+### Removed
+
+- Individual scripts
+
+### Fixed
+
+- Fixed various bugs with parsing annotations from EPPIJson
+
+## [0.2.0] - 2025-12-11
+
+### Added
+
+- Data models
+- EPPIJson parser
+- Data Extractor class
+- Pipelines
+
+### Changed
+
+- Renamed main package directory from `app` to `deet`
+- Made `deet` installable
+
+### Removed
+
+## [0.1.0] - 2025-09-03
+
+## Added
+
+- Initial project scaffolding
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Evaluation of data extracted by LLMs
 - Documentation
 - Parsing of annotation data in CSV format
+- Dynamic versioning
 
 ### Changed
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,13 +2,20 @@
 
 Last updated: NL, 2025-02-19
 
-## Generally applicable guidelines
+## Development workflow
 
 - Any contribution should be encapsulated within a pull request (PR), from a new branch whose sole purpose is the implementation of the contribution.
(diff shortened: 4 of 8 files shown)


mcallaghan:
reviewers, when we are ready to merge - let's add the date to the proposed 0.3.0 release. After merging we tag this release.

mcallaghan:
## Technical ways to implement this

We discussed doing this with uv before. Most of this would work in a similar way except we replace 
```sh
git tag 0.3.0
git push --tags
``` 
with 
```sh
uv version 0.3.0 # OR uv version --bump patch
git add pyproject.toml
git commit -m "bump version 0.3.0"
git push
```

If we want the uv route, then we just revert all changes to project.toml and uv sync again

The main difference between the two approaches is that with uv, the version will always be 0.3.0, regardless of what is committed, until the next time the version is bumped. With setuptools_scm, a string is automatically appended indicating the number of commits since the last manual version, and a commit hash.

mcallaghan:
## Organisational Questions

Aside from technical questions. There are a few choices about how to set up the process. Bumping the version only when we merge dev into main is probably low maintenance. If we want a bit more granularity (especially in the uv route), we can bump the version each time we merge into dev.

.gitignore
```diff
@@ -224,3 +224,6 @@ default_extraction_config.yaml
 pdfs/*
 data-extraction-experiments/*
 *.csv
+
+# Dynamic version
+deet/_version.py
```
nikloynes: should this be called `.version.py` if goal is for it to be hidden? what's the use case of this that isn't in pyproject?

.markdownlint.yaml
```diff
@@ -4,3 +4,5 @@ MD013: false #  line-length
 ignore:
   - "tests/test_files/**/*"
 MD046: false # code block style (need both styles if tabs are to work)
+MD024:
```
nikloynes: what does this do?

CHANGELOG.md
nikloynes, line 1: looks good. IMO no need to backfill. is there potentially some sort of method/tool that also lets us auto-add commit messages here, or link them at least? that might be useful.

CONTRIBUTING.md
```diff
@@ -2,24 +2,39 @@
...
   - all existing and new tests are passing, both locally and in Continuous integration (CI)
   - the core functionality of the application (i.e. the core CLI data extraction flow) is still functional locally (as we currently don't test this in CI)
   - your contribution passes linting (`ruff`) and `mypy`.
   - your contribution is well-documented, to the point that the PR summary itemises the changes you've made.
+  - __Changelog__: You have added a descriptive bullet point to the `[Unreleased]` section of `CHANGELOG.md`.
 - Note that you can't expect your colleagues to include running your code in the context of reviewing it. __The onus of ensuring a) that your code works and b) that it doesn't break existing functionality is ___on you___.__
 - Copilot can be a decent PR reviewer, especially before you ask a fellow contributor for a review. Copilot alone should typically not be sufficient for allowing a PR to be merged however.
 - Once a PR is approved and ready for review, the original author should merge the commit into the target branch.
 
+### Automated Versioning
```
nikloynes, line 31: not sure i'm a huge fan of this, find path, minor, major easier to follow and having that as a component of a PR.

CONTRIBUTING.md
```diff
@@ -2,24 +2,39 @@
 
 Last updated: NL, 2025-02-19
 
-## Generally applicable guidelines
+## Development workflow
```
nikloynes, line 5: seems like this has quite a different meaning from what it replaces

pyproject.toml
```diff
@@ -50,10 +51,10 @@ dependencies = [
     "typer>=0.21.2",
 ]
 description = "Data Extraction Evaluation Toolkit (DEET): A suite of tools, data models, etc. for extracting data from documents (e.g. papers) and evaluating the performance of such extraction tasks"
+dynamic = ["version"]
```
nikloynes, line 54: so version is no longer human-readable in pyproject?

nikloynes (commented):
scaffolding & changelog setup makes sense and looks good to go. i find `uv version --bump minor` much more intuitive and a good final step in a PR. i like being able to read version number in pyproject. 

EDIT: also, shouldn't this PR now point to dev first?

.gitignore
mcallaghan: This is actually no longer necessary in python3.12, so I've removed this, good question!

.markdownlint.yaml
mcallaghan: I've added a comment. Basically allows multiple sections with the same heading, as long as they are not siblings. This is necessary for the changelog, otherwise it will complain at the repetition of the same sections (fixed, added, deprecated etc.) in multiple versions

CHANGELOG.md
mcallaghan: Yes, I thought about this, but for this to work well, we need to have more disciplined structured commit messages. It may also be that our commits have a level of detail that we don't actually require in the changelog, and may be addressing different audiences. I thought on balance it might be easier to not automate this.

pyproject.toml
mcallaghan: no, it is not - we are basically replacing a hard-coded variable with a function that calculates the version based on git commits.

CONTRIBUTING.md
mcallaghan: Yes, though I'm not trying to suggest anything below is no longer generally applicable - just wanted to introduce some sections as the text gets longer, and the subheadings made more sense to me under this title, otherwise it reads to me as if only the first two points are generally applicable. Happy to revert

CONTRIBUTING.md
mcallaghan: We still do patch, minor, and major, but we do this only when consolidating dev into main (or I suppose there might be a case where we patch directly to main). Only the bit after the normal semantic version is automatically generated. The normal semantic version is defined manually by tagging a commit.

PR #244, opened 2026-04-24
Feature/241 extraction without linking or evaluating

### LINKED ISSUE #241: Extract data without gold standard annotation data
### Description

Sometimes we want to extract data from a document and compare it to gold standard data, but sometimes we just want to extract data, e.g. because we do not have gold standard data. For this, we want a command along the lines of `deet predict`, that runs the usual pipeline, but doesn't try any evaluation

### Definition of Done

- [ ] Command in the CLI to run extraction
- [ ] Documentation in the docs of how to do this

### Comments

I assume that we still want to define docs and attributes in a csv / EPPIJSON, but we just ignore their annotations.

mcallaghan:
# Pull Request

## Description
- Adds a new command to generate predictions without evaluating: `deet experiments predict`. The results are written to csv.
  - Adds the option `--ignore-references` to extract from all documents in the pdf folder, without attempting to link these to bibliographic records 
- Renames `deet experiments extract` to `deet experiments evaluate` to clearly separate the two
- Shifts main extraction logic into a helper function which is run by the two commands

## Related Issue(s)
Closes #241 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Testing
Interactive testing with `SmokingTest.json`

To extract without li

```sh
deet project init
# Edit csv 
```

To extract without linking:

```sh
deet experiments predict --ignore-references
````

Or to extract linked documents
```sh
edit link map
deet project link
deet experiments predict # Runs data without evaluation
deet experiments evaluate # Runs data and evaluates
```

## Additional Notes
This outputs the standard json annotated documents, as well as a csv with the LLM annotation for each document-attribute combination.

---


AilbheF:
Points to the interactive cli which is going into development. Converting to csv introduces bugs so about how do we want to visualise the output, currently annotated_docs.json

Need a list of attributes and list of documents. EPPI.json as the import - codeset without the references

Future work - just have a folder of pdfs and a list of data extractions terms with no linking required for the pdfs as there is no associated gold standard data to link to.

deet/scripts/commands/experiments.py
```diff
@@ -24,7 +19,7 @@
 
 @app.command()
 @project_required
-def extract(
+def evaluate(
```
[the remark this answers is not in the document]
mcallaghan: Since extract was never merged into dev, I don't think we need to maintain backward compatibility

AilbheF (approved):
works as intended with eppi json data codeset without any gold standard data for boolean annotations. Did not try with string attributes. 

Only error was the inability to increase max context tokens even though edited in config file - similar error encountered previously - will make a bug issue for this

deet/data_models/documents.py
```diff
@@ -539,16 +539,19 @@ def get_attribute_annotation(self, attribute: Attribute) -> GoldStandardAnnotati
...
-            except ValueError as err:
-                not_found = (
-                    "Attribute not found in annotations."
-                    " Don't know how to interpret this when attribute is of type "
-                    f"{attribute.output_data_type}"
+                reasoning = (
+                    "No annotation with this attribute found: reverting to default"
+                )
+            except ValueError:
+                output_data = ""
```
mcallaghan, line 546: Instead of raising an error (which we don't like to see), we just put a null value and explanation into wherever we end up writing this

deet/data_models/project.py
```diff
@@ -248,3 +248,8 @@ def config_snapshot(self) -> Path:
     def llm_annotations(self) -> Path:
         """Return location of json containing llm extractions."""
         return self.base_dir / "llm_annotations.json"
+
+    @property
+    def llm_annotation_csv(self) -> Path:
+        """Return location of csv containing llm extractions."""
+        return self.base_dir / "llm_annotations.csv"
```
mcallaghan, line 254: This is a new csv shaped like the human-llm comparison csv, but it only contains llm annotations and reasoning

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -401,3 +401,54 @@ def export_llm_comparison(
                             "extraction_run_id": self.extraction_run_id,
                         }
                     )
+
+    def export_llm_csv(self, filepath: Path) -> None:
+        """Write the LLM output to csv."""
```
mcallaghan, line 406: This is a new function to write the aforementioned csv

deet/extractors/cli_helpers.py
```diff
@@ -137,3 +144,70 @@ def prepare_documents(
         fail_with_message(message)
 
     return None
+
+
+def run_extraction_pipeline(
```
mcallaghan, line 149: Code from the CLI command is now moved into this helper function, as extraction is now done in two separate commands

deet/processors/directory_processor.py
```diff
@@ -0,0 +1,50 @@
+"""Tools to create Documents directly from pdf or md files."""
```
mcallaghan, line 1: In order to implement extraction without linking, we use this module to create "fake" linked documents from a directory.

deet/processors/directory_processor.py
```diff
@@ -0,0 +1,50 @@
...
+
+    mock_references = []
+    mock_mappings = []
+
+    for file_path in target_files:
+        document_id = hash_n_strings_to_document_id([file_path.stem])
+
+        mock_ref = Document(
+            name=file_path.name,
+            document_id=document_id,
```
mcallaghan, line 30: As there is no gold standard document to link to, we just create a fake document with an ID produced by hashing the file path. The name is the file_path, so it is easy for the user to tie these documents back to the files.

deet/processors/directory_processor.py
```diff
@@ -0,0 +1,50 @@
...
+
+    if not mock_references:
+        return []
+
+    linker = DocumentReferenceLinker(
+        references=mock_references,
+        document_reference_mapping=mock_mappings,
+        document_base_dir=directory_path,
+    )
+    return linker.link_many_references_parsed_documents()
```
mcallaghan, line 50: We then create these "linked documents" in the same way as they would usually be created, by connecting a list of document objects to a list of file paths, and letting the standard `DocumentReferenceLinker` handle parsing etc.

mcallaghan (commented):
Added some comments to explain changes...

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -401,3 +401,54 @@ def export_llm_comparison(
...
+        with filepath.open("w", newline="", encoding="utf-8") as f:
+            writer = csv.DictWriter(
+                f,
+                fieldnames=[
+                    "document_id",
+                    "document_name",
+                    "attribute_id",
```
nikloynes, line 413: so where does it take the att id & label from?

deet/processors/directory_processor.py
```diff
@@ -0,0 +1,50 @@
...
+    """Find PDF and Markdown files turns them into linked documents."""
+    target_files = list(directory_path.glob("*.md"))
+    md_stems = {p.stem for p in target_files}
+
+    target_files.extend(
+        p for p in directory_path.glob("*.pdf") if p.stem not in md_stems
+    )
+
+    mock_references = []
+    mock_mappings = []
```
nikloynes, line 23: can't we simplify this, globbing for md and pdf simultaneously?

deet/processors/directory_processor.py
```diff
@@ -0,0 +1,50 @@
...
+
+    for file_path in target_files:
+        document_id = hash_n_strings_to_document_id([file_path.stem])
```
nikloynes, line 26: this should probably be implemented as a linking strategy then so that extractions from this here are forwards compatible with deet runs.

deet/processors/directory_processor.py
nikloynes, line 1: this is clever. just wonder if this belongs in a helper/utils folder rather than here?

nikloynes (approved):
looks good. some minor commentd here and there.

deet/processors/directory_processor.py
mcallaghan: the reason for this is to prefer md to pdf if both exist. Otherwise if both exist we'd create a document for each.

deet/evaluators/gold_standard_llm_evaluator.py
mcallaghan: we still need to provide a gold standard data path when creating a project, it just doesn't need to contain any documents (just attributes). This can be as simple as a csv with just headers, where each column is an attribute

mcallaghan: or it could contain documents, but we just ignore these when running predict with the --ignore-references, because we don't care at this point about linking the documents in the gold standard data to their pdfs

deet/processors/directory_processor.py
mcallaghan: Ohhh, I think I see a way to do this, this whole module could go into linker as a strategy. And that strategy could operate *only* when the list of references is empty, and just implement this. Perhaps we can leave this to a later date - as I think this would imply some more changes to linker than I think we need to touch now. But that would be a clean way to do this

deet/processors/directory_processor.py
mcallaghan: let's leave it in here pending subsequent integration as a strategy all in itself (see above)

deet/processors/directory_processor.py
mcallaghan: #282

deet/evaluators/gold_standard_llm_evaluator.py
AilbheF: I might need to try this again but when exporting a CodeSet from EPPI as a json you have to include at least one reference. It did not work when I used that json file (with one reference) until I had removed the data and just left the CodeSet. The error might have been related to something else so I can double check to be sure

mcallaghan: Do you want to send me the CodeSet? Or paste in here? Happy to have a look

Issue #233, opened 2026-04-20
Develop full feature specification for arm extraction


kaitlynhair:
### Description
The ability to extract experimental arms (i.e. experimental groups) is the next key milestone we want to achieve in DEET. This is complex and may require changes to the data input/output schemas and evaluation strategies. 

### Definition of Done

- [ ] In person meeting (UCL) to discuss and draft initial feature spec with detailed exemplars and use cases
- [ ] Consultation with wider WP2/3 team on feature spec 
- [ ] Technical implementation plan developed
- [ ] GitHub issues added and roles assigned

### Dependencies

This relates to the implementation of list data types #232

kaitlynhair:
UCL research team hoping to meet this week to discuss and develop initial feature spec to present at next Thursday meeting

mcallaghan:
We should consider parallel work on data standards, see https://gist.github.com/mootpointer/3c3cd5fa6b2e11fda5c58d99db07b024#file-evidence-repository-technical-docs-md

mcallaghan:
> [!WARNING]
> When we address this, we should make sure to adjust annotation deduplication, see #246

Issue #231, opened 2026-04-20
Implement integration tests for DEET


kaitlynhair:
### Description
To ensure DEET successfully runs on different datasets when updates are made, we need integration tests which send data to the LLM, retrieve output, produce valid output, and produce valid evaluation files. 

### Definition of Done

- [ ] live extraction tests for each available extraction/output type parameterised by single sample from each core dataset
- [ ] eval tests parameterised by single sample from each core dataset
- [ ] add windows type runners 
- [ ] start with EPPI JSON, expand to CSV imports down the line

### Dependencies

- #Issue

### Comments

We need to clarify the "core datasets". Is it just the sub-folders of the Datasets folder in DESTINY OneDrive?

kaitlynhair:
Defining what we want to test: extracting an attribute of each type, evaluation for each type.
Use datasets in sharepoint and take small OA sample from each to facilitate testing. 

- [ ] Each person who has a dataset (bool + string, just bool, just string) to create a small EPPI JSON with 1/2 open access records and associated documents and CSV

PR #246, opened 2026-04-27
Feature/232 lists

### LINKED ISSUE #232: Implement list data type for DEET imports
### Description

We should use lists whenever a piece of data can have 0 or more instances.

For example, given the prompt "extract all mentions of countries from the study", the following answers would all be valid

```python
[] # Nothing, there were no countries
["Germany"] # A single country was mentioned
["Germany", "France"] # More than 1 country was mentioned
```

We already have list-like data in EPPIJson imports, but this is implemented with multiple items in the codeset of a document with the same attribute_id.

Note: In the data I was looking at (SmokingTest), this was actually accounted for by the fact that the duplicated text referred to different arms, but this is a separate issue (#233)

To achieve this we need to change eppiannotationconverter to collapse multiple codes with the same attribute into one annotation with a value for raw_data that separates values with a separator (e.g. ;;;), then write a coercion strategy to `deet.data_models.base.ANNOTATION_COERCION_STRATEGIES`, that separates these into a list, when we specify that the attribute is of type list.

### Definition of Done

- [ ] Our EppiJSON converter converts duplicate annotations into a single annotation with all values separated by a separator in `raw_data` (see #179) 
- [ ] A coercion strategy separates raw_data into a list

### Dependencies

- #234 

### Comments

After we've implemented this, we need to figure out evaluation metrics for lists

mcallaghan:
# Pull Request

## Description
Add support for lists by merging duplicated eppi annotations, and adding a list coercion strategy to output_data

### Deduping and merging
Adds the functions `dedup_annotations` and `_merge_raw_values()` to `deet.processors.eppi_annotation_converter.py`, which identify duplicate annotations and merge their raw values. Values Are merged into strings separated by ";;; "s, unless they are alreeady a list, in which case the list is extended, or they are dict, in which case they are merged.

> [!WARNING]
> Currently, annotations are considered duplicates if they share an attribute ID and a document ID. 
> This will need to be adjusted when we support arms #233 

### Coercing to list
In `deet.data_models.base` add `coerce_annotation_to_list` to `ANNOTATION_COERCION_STRATEGIES`.

## Related Issue(s)
Closes #232 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Testing
- `tests/unit/test_base.test_list_coercion` tests that list coercion works as intended for a variety of cases
- `tests/unit/test_base.test_gold_standard_annotation_conversion_to_list` tests that this works on a real annotation
- `tests/unit/test_eppi_annotation_converter.test_process_annotation_file_with_duplicated_annotations` tests that annotations are properly deduped and merged

## Additional Notes

---


AilbheF:
Sneaking it in before versioning and locking down main before next release. Important to merge as it impacts several of the test datasets

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+def coerce_annotation_to_str(val: SUPPORTED_TYPES) -> str:
+    """Coerce an annotation to a string."""
+    return str(val) if val else ""
+
+
+def coerce_annotation_to_bool(val: SUPPORTED_TYPES) -> bool:
+    """Coerce an annotation to a bool."""
+    if isinstance(val, bool):
+        return val
+    if isinstance(val, str) and val.lower() in ("false", "0"):
```
sagaruprety, line 260: A numeric 0.0 coerces correctly to False, but "0.0" will be True because it falls into the string path

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
 AttributeTypeVar = TypeVar("AttributeTypeVar", bound=Attribute)
 
 
+def coerce_annotation_to_str(val: SUPPORTED_TYPES) -> str:
+    """Coerce an annotation to a string."""
```
sagaruprety, line 252: if val treats 0, 0.0, False, and [] as falsy; we end up returning "" instead of "0" or "False". What if we do simple this : 
```

def coerce_annotation_to_str(val: SUPPORTED_TYPES | None) -> str:
    if val is None:
        return ""
    return str(val)
```

This preserves 0, False, and empty containers as meaningful strings.

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+        try:
+            return float(val)
+        except ValueError:
+            logger.warning(f"Could not convert {val} to float")
+
+    logger.warning(f"Unsupported type for float conversion: {type(val).__name__}")
+    return None
+
+
+def coerce_annotation_to_list(val: SUPPORTED_TYPES) -> list:
```
sagaruprety, line 297: maybe add a short comment/docstring explaining the ';;;' delimiter

deet/data_models/base.py
VJonah: Good point. I'd rather we didn't accept None as input and simply test for empty collections i.e. dict and list then return `str(val)`:

```python
if isinstance(val, dict) and not val:
    return ""
if isinstance(val, list) and not val:
    return ""
return str(val)
```

That sort of thing?

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+    if isinstance(val, str) and val.lower() in ("false", "0"):
+        return False
+
+    if isinstance(val, int | float):
+        return bool(val)
+
+    return True
+
+
+def coerce_annotation_to_int(val: SUPPORTED_TYPES) -> int | None:
```
VJonah, line 269: I think it might be worth seeing if we can avoid returning `None` here. If we're going to have these coercion functions, I think it's worth them sharing the same type interface (i.e. `SUPPORTED_TYPES` -> concrete type)

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+        try:
+            return int(val)
+        except ValueError:
+            logger.warning("Could not convert {val} to int")
+
+    logger.warning(f"Unsupported type for int conversion: {type(val).__name__}")
+    return None
+
+
+def coerce_annotation_to_float(val: SUPPORTED_TYPES) -> float | None:
```
VJonah, line 283: Same as above. Thinking about it, we're also just implicitly failing on our collections. It might be worth having an exception like `CannotCoerceCollectionToNumbericTypeException` and propagate them.

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+        except ValueError:
+            logger.warning(f"Could not convert {val} to float")
+
+    logger.warning(f"Unsupported type for float conversion: {type(val).__name__}")
+    return None
+
+
+def coerce_annotation_to_list(val: SUPPORTED_TYPES) -> list:
+    """Coerce an annotation to list."""
+    if val is None:
```
VJonah, line 299: Why would `val` ever be `None` if the function only takes a non-None `SUPPORTED_TYPES` arg

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+        return []
+    if isinstance(val, list):
+        return val
+    if isinstance(val, str):
+        list_of_strings = [item.strip() for item in val.split(";;;")]
+        try:
+            return [int(item) for item in list_of_strings]
```
VJonah, line 306: Why not reuse the `coerce_annotation_to_X` functions here? If we expect nested subtypes, why not recursively apply our own coercion logic? I think this is where enforcing a uniform interface and raising custom exceptions as above instead of `None`, would be of benefit.

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+        except (ValueError, TypeError):
+            logger.debug("Could not convert items to int.")
+
+        try:
+            return [float(item) for item in list_of_strings]
```
VJonah, line 311: Same here.

deet/data_models/base.py
```diff
@@ -246,11 +248,101 @@ def sanitize_prompt(prompt: str) -> str:
...
+}
+
+
 class GoldStandardAnnotation(BaseModel):
-    """A single gold standard annotation for an attribute."""
+    """
+    A single gold standard annotation for an attribute.
+
+    `raw_data` stores the data as it comes from source,
+    `output_data` is computed and coerces raw_data into the correct type.
```
VJonah, line 335: general question: what dictates what the "correct type" is?

deet/data_models/base.py
```diff
@@ -263,25 +355,22 @@ class GoldStandardAnnotation(BaseModel):
 
     @model_validator(mode="before")
     @classmethod
-    def ensure_correct_type(cls, data: dict) -> dict:
-        """Ensure output_data is of the type required by annotation_type."""
-        target_att: Attribute = data["attribute"]
+    def handle_output_data_input(cls, data: SUPPORTED_TYPES) -> SUPPORTED_TYPES:
```
VJonah, line 358: I'm not sure I understand what this does, sorry?

tests/unit/test_base.py
```diff
@@ -786,12 +788,84 @@ def test_gold_standard_annotation_bool_type_invalid() -> None:
...
+        attribute=attr,
+        raw_data="not a bool",
+        annotation_type=AnnotationType.HUMAN,
+    )
+
+    assert isinstance(annotation.output_data, bool)
+    assert annotation.output_data
+
+
+def test_gold_standard_annotation_string_preserved_on_type_change() -> None:
```
VJonah, line 801: general question: why would this change, sorry?

VJonah (approved):
A few small things on the coercion functions. I'm not entirely clear on the coupling of attributes to annotations like GoldStandardAnnotation. But if this solves list extraction for people who need it quickly. Then happy to merge into main, lock the branch, and revisit the annotation-attribute design in future development.

deet/data_models/base.py
mcallaghan: ah sorry, this was pointing to dev, after I changed this during our meeting, but I didn't change back to main after our discussion. This is already in main

deet/data_models/base.py
mcallaghan: see above, already in main, but PR had been pointed to dev

deet/data_models/base.py
mcallaghan: I'd like this too, but how can we int "five", or "python". Outputting None means everything can continue, but the metrics won't make sense. In any case, this is, as above, already in main

deet/data_models/base.py
mcallaghan: see above - something for the next iteration!

tests/unit/test_base.py
mcallaghan: The EPPIAnnotationConverter parses everything as Bool. But when you set the type with the prompt csv, it changes the AttributeType of the already created bool Attributes. Previously, this was not doing anything to the annotations. We were wondering last week if we'd have to re-parse the whole annotation data after setting the types manually in the prompt csv, but by implementing these coercion strategies and making output_data a computed_field, we can change the type of already existing attributes and not have to re-parse the data

deet/data_models/base.py
VJonah: In my opinion, returning `None` just propagates checks to the caller without providing any details/semantics as to why coercion failed. A custom exception could address this. It might also be worth prompting the user for corrections on failed coercions.

deet/data_models/base.py
nikloynes: do you mean 'iteratively' or 'recursively' here, @VJonah ?

deet/data_models/base.py
```diff
@@ -294,13 +294,36 @@ def coerce_annotation_to_float(val: SUPPORTED_TYPES) -> float | None:
...
 
 
+def coerce_annotation_to_list(val: SUPPORTED_TYPES) -> list:
+    """Coerce an annotation to list."""
+    if val is None:
+        return []
+    if isinstance(val, list):
+        return val
+    if isinstance(val, str):
+        list_of_strings = [item.strip() for item in val.split(";;;")]
```
nikloynes, line 304: IMO `;;;` should
- be a constant at the top, perhaps `DEFAULT_ARRAY_SEPARATOR`
- an arg to the function, so that different array separators can also be passed

deet/data_models/base.py
```diff
@@ -294,13 +294,36 @@ def coerce_annotation_to_float(val: SUPPORTED_TYPES) -> float | None:
...
+        try:
+            return [int(item) for item in list_of_strings]
+        except (ValueError, TypeError):
+            logger.debug("Could not convert items to int.")
+
+        try:
+            return [float(item) for item in list_of_strings]
+        except (ValueError, TypeError):
+            logger.debug("Could not convert items to float")
+
```
nikloynes, line 314: if it fails both float and int, should it perhaps throw an error we can catch elsewhere?

deet/data_models/base.py
```diff
@@ -294,13 +294,36 @@ def coerce_annotation_to_float(val: SUPPORTED_TYPES) -> float | None:
...
+        return list_of_strings
+    return [val]
+
+
 ANNOTATION_COERCION_STRATEGIES: dict[
     AttributeType, Callable[[SUPPORTED_TYPES], SUPPORTED_TYPES | None]
```
nikloynes, line 320: @VJonah this is presumably why max has included `None` in the list coercion func

deet/data_models/base.py
nikloynes: i think this will be passed by the user, no?

deet/data_models/base.py
nikloynes, line 1: i realise this might be a little older, but not quite sure why we're no longer using the OOP style in this addition (the coercion bits) and opting for a functional approach. obviously doesn't mean it won't work, seems inconsistent, and might also take a few more lines. otherwise seems to make sense besides my points on letting the user pass the sep char(s) and having the ;;; as a constant not a hidden default.

deet/processors/eppi_annotation_converter.py
```diff
@@ -270,6 +270,21 @@ def _convert_single_annotation(
...
+    ) -> SUPPORTED_TYPES:
+        """Merge the raw_data of duplicated annotations."""
+        if existing is None:
+            return new
+        if new is None:
+            return existing
+        if isinstance(existing, list) and isinstance(new, list):
+            return existing + new
+        if isinstance(existing, dict) and isinstance(new, dict):
+            return {**existing, **new}
```
nikloynes, line 284: do we actually ever have dict data?

deet/processors/eppi_annotation_converter.py
```diff
@@ -302,6 +317,29 @@ def convert_to_eppi_annotations(
                 continue
         return results
 
+    def dedup_annotations(
```
nikloynes, line 320: should be `dedupe_annotations`
```suggestion
    def dedupe_annotations(
```

nikloynes (approved):
mostly good to go, but i haven't used list data to test it, so that should take precedence in deciding whether it should get merged.

deet/data_models/base.py
VJonah: Is there a specific reason we use ';;;' as delimiter you want mentioned? I'm not sure I'm familiar with the reason of using ';;;' specificlally.

deet/data_models/base.py
VJonah: 'recursively'. As in, we can recursively call our coercion functions nested types. e.g. if we have a list of lists of str, we can call our coercion functions within themselves, i.e. mutual recursion. Even if we don't expect it, I think we should still call our own coercion functions 'iteratively' when constructing list objects

deet/data_models/base.py
VJonah: How so? The type hint is just saying that an Attribute type can map to a function that returns either a `SUPPORTED_TYPES` instance or `None`. I think it's the other way around, no? Max created functions that returned `None` and so put the appropriate type hint here.

deet/data_models/base.py
AilbheF: This is just from the dataset that I have which was the data used for the original version of DEET and we needed to separate spans of text in the info box in eppi and needed something that would not be a natural occurring separator and this ;;; was what was suggested. It is not a requirement it is just something we had used previously but if there is something better feel free to change. 

Also this worked in an example I shared on the teams chat where sometimes it used " " as a separator or text and other times it is ;;; it might be depending on their location on the page in the PDF or when they are on separate pages

AilbheF: Actually it might be when they are different arms rather than where they are in the text - will double check the raw data

VJonah:
As @mcallaghan mentioned, a few of these comments point to code already in main because of switching the branch the PR points to. Ultimately, the net changes are relatively small. It's worked on a given dataset. So it's not necessarily worth refactoring main and this PR in anticipation of future refactors brought about by integration tests: #231

Issue #193, opened 2026-03-23
Explore addition of "verbatim text" field back into data model


kaitlynhair:
### Description
LLM response has reasoning field + VERBATIM text field (text used to make decision). We need to discuss how to add this into the data models. In original versions of the code but removed at some stage. 

Aim of the field is to 1) use LLM to pre-extract data to support people doing systematic reviews (highlighted text) or 2) to support validation of LLM-based data extraction vs gold standard data. 

### Definition of Done

- [ ] Verbatim text from PDF is in LLM output 
- [ ] Verbatim text matches to source context (hallucination check would require a text match in PDF)
- [ ] Determine how failure cases would be handled for the above (maybe it doesn't return a match if no fuzzy match e.g. Langextract)

### Dependencies

- #Issue

### Comments

- Need to ensure we distinguish between when LLM not accurate for this (extracts wrong text) vs function not working as expected 
- What happens when the source context doesn't match 100% but is a partial match
- Need to be aware of inconsistencies in parsing from PDF -> markdown and this may cause discrepancies vs the original dataset 
- Degree of hallucination may need to be quantified (e.g. hyphen missing vs entire words missing) 
- Could retrieve indices (from markdown) rather than specific extracted text (see Langextract) - can then extract the text between these indices and may reduce the risk of LLMs generating from text - needs experimentation to determine best approach to reduce hallucination errors

PR #248, opened 2026-04-28
Fix/218 verbatim text output

### LINKED ISSUE #123: note for re-building the extract_from_documents() for batching
presumably, once we switch to batching, we should actually populate this method. 
it would also probably make sense to have `extract_from_document` yield a generator that we can then iterate over here. @sagaruprety is that how you'd thought of/implemented it?

_Originally posted by @nikloynes in https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/pull/121#discussion_r2709562568_

### LINKED ISSUE #218: EPPI-json source of gold standard text
### Description
EPPI jsons can have gold standard text data in two places in the reference code annotations:
1. "AdditionalText" which is the text users fill into the info box in EPPI Reviewer
2. "Text" content from dicts in the lists contained in "ItemAttributeFullTextDetails"

Previously, "AdditionalText" was default but recently, "AdditionalText" content was not outputted anymore into the comparison CSVs. Instead, the often empty direct "Text" annotations are used.

We required the  "AdditionalText" for the student project, which caused a problem because the data were not in the output.

As quick fix, I implemented pulling data from both fields within [eppi-ingestion-processing](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/tree/eppi-ingestion-processing)

Chnages needed for this are in [this commit](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/commit/73676d4b220546a5e9f44ca68141111a96e93039) in files deet/evaluators/gold_standard_llm_evaluator.py and [eppi annotation converter](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/8be0383ee0488994408e58b547497e26bb91bb1a/deet/processors/eppi_annotation_converter.py#L265-L273)

Options to deal with this:
- Implemented fix: Pulling and returning both (i.e. having two gold standard columns) leads to the most flexibility, but would then require the user to specify which column to use for downstream evaluation.
- Decide on one to be default, and allow user to specify and change this using an argument. If default is empty, then no annotations are returned
- Decide on one to be default, but then pull data from the other one if empty (Note, I think this might lead to confusion, issues, or bugs)

### Definition of Done

- [ ] Consensus on default field and implementation is reached
- [ ] Data from @AilbheF datasets with text annotations as well as data from @L-ENA and @kaitlynhair can be displayed in the CSV output

sagaruprety:
# Pull Request

## Description
Fixes #218 and #193. 
Added new fields in the output file goldstandard_llm_comparison.csv — attribute_presence (bool, whether an attribute is present in the ground truth or not, regardless of its output type), human_additional_text (the additional_text field in eppi json), item_attribute_full_text_details (the namesake field in eppi json), llm_verbatim_text (llm equivalent of additional_text), human_verbatim_fuzzy_match_pct / llm_verbatim_fuzzy_match_pct (how much of the respective fields match the markdown text); 

EPPI ingest: eppi_output_data_from_eppi_fields maps AdditionalText → typed raw_data (non boolean); populate_custom_prompts on EPPI data recomputes raw_data after prompt/type CSV load
s.

Tests: Comparison CSV header / presence checks in test_gold_standard_llm_evaluator

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Pre-Review Checklist

- [ ] Code passes `mypy` type checking

## Additional Notes

---


deet/data_models/processed_gold_standard_annotations.py
```diff
@@ -359,3 +359,37 @@ class ProcessedEppiAnnotationData(
...
+    def populate_custom_prompts(
+        self,
+        method: CustomPromptPopulationMethod,
+        filepath: Path | None = None,
+        **kwargs: Any,  # noqa: ANN401
+    ) -> None:
+        """
+        Populate custom prompts, then recompute each annotation's ``raw_data``.
+
+        An EPPI ``Report.json`` (or similar) only carries EPPI-Reviewer field names;
```
nikloynes, line 372: so by this you mean an eppi json?

deet/data_models/processed_gold_standard_annotations.py
```diff
@@ -359,3 +359,37 @@ class ProcessedEppiAnnotationData(
...
+        ``eppi_output_data_from_eppi_fields`` so ``raw_data`` matches
+        ``AdditionalText`` and the updated type (e.g. ``"32"`` for a string count,
+        not boolean ``True`` from initial ingest only).
+        """
+        super().populate_custom_prompts(method, filepath=filepath, **kwargs)
+        self._recompute_eppi_raw_data_from_additional_text()
+
+    def _recompute_eppi_raw_data_from_additional_text(self) -> None:
+        """Re-derive each gold ``raw_data`` from types and ``AdditionalText``."""
+        from deet.processors.eppi_annotation_converter import (
```
nikloynes, line 390: are we relying on the cheaper lazy import here or could it go to the top?

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -28,6 +30,57 @@
 )
 from deet.exceptions import DuplicateAnnotationError, MissingDocumentError
 
+# Use stricter heuristics (substring / digit boundary) for very short verbatims.
+_VERBATIM_FUZZ_SHORT_LEN = 4
```
nikloynes: unclear what this constant means/does, maybe add a few words in comment to explain

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -28,6 +30,57 @@
...
+
+
+def _verbatim_fuzzy_match_pct(verbatim: str | None, context: str | None) -> float:
```
nikloynes: could we re-work the arg names? i find them not intuitive. what _exactly_ are we comparing?

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -28,6 +30,57 @@
...
+    Args:
+        verbatim: Human or model snippet (e.g. ``additional_text``).
+        context: Document text used for comparison.
+
+    Returns:
+        Float in ``[0.0, 100.0]``, or ``0.0`` if either input is empty.
+
+    """
+    v = (verbatim or "").strip()
+    c = (context or "").strip()
```
nikloynes: this and following now mean a not very easily readable/understandable algorithm.

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -28,6 +30,57 @@
...
+    if not v or not c:
+        return 0.0
+    if len(v) < _VERBATIM_FUZZ_SHORT_LEN and v.isdecimal():
```
nikloynes: still unclear what this constant does, but might be good to add it as an arg to the func and set as default.

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -28,6 +30,57 @@
...
+            return 100.0
+        return 0.0
+    if len(v) < _VERBATIM_FUZZ_SHORT_LEN:
+        return 100.0 if v in c else float(fuzz.partial_ratio(v, c))
+    return float(fuzz.partial_ratio(v, c))
+
+
+def _eppi_full_text_details_colon_separated(annotation: object) -> str:
+    """
+    Colon-join all non-empty ``Text`` values from ``item_attribute_full_text_details``.
```
nikloynes: why colon?

deet/evaluators/gold_standard_llm_evaluator.py
```diff
@@ -191,7 +244,24 @@ def export_llm_comparison(
...
+        """
+        Export a comparison CSV for gold vs LLM per document and attribute.
+
+        Columns include identifiers, EPPI-oriented fields, extractions, LLM verbatim,
+        fuzzy grounding scores (against the LLM annotated document's ``context``),
+        and run id.
+
+        Column semantics:
+
+        - ``attribute_presence``: Whether the gold annotation is present.
```
nikloynes, line 287: would be good to add example entry/data type to give reader more context here

deet/processors/eppi_annotation_converter.py
```diff
@@ -29,6 +29,72 @@
     Outfiles,
 )
 
+EppiRawDataValue = bool | str | int | float | list[Any] | dict[str, Any]
```
nikloynes: if a constant, it should be all caps. but i'm also not quite sure what this is supposed to be? all possible ways that we ingest raw data from eppi? are there any types we haven't noted here? i guess None.

deet/processors/eppi_annotation_converter.py
```diff
@@ -29,6 +29,72 @@
...
+
+
+def eppi_output_data_from_eppi_fields(  # noqa: PLR0911
+    output_data_type: AttributeType,
+    *,
+    additional_text: str,
+) -> EppiRawDataValue:
+    """
+    Map EPPI ``Codes`` row evidence onto typed ``raw_data`` for coerced ``output_data``.
```
nikloynes: would be great if you can explain what `Codes` are, as well as `raw_data` and `output_data` in this context. too many context-specific terms are getting me confused...

deet/processors/eppi_annotation_converter.py
```diff
@@ -29,6 +29,72 @@
...
+
+    Returns:
+        Value to store in ``GoldStandardAnnotation.raw_data`` (then coerced via
+        ``output_data``).
+
+    """
+    additional = (additional_text or "").strip()
+
+    if output_data_type == AttributeType.BOOL:
+        return True
```
nikloynes, line 149: you'll have to explain why this defaults to true

deet/processors/eppi_annotation_converter.py
```diff
@@ -29,6 +29,72 @@
...
+    if output_data_type == AttributeType.STRING:
+        return additional
+
+    if output_data_type == AttributeType.INTEGER:
+        if not additional:
+            return output_data_type.missing_annotation_default()
+        try:
+            return int(float(additional))
+        except ValueError:
```
nikloynes: is there a world where we should add a match clause here to specify?

deet/processors/eppi_annotation_converter.py
```diff
@@ -29,6 +29,72 @@
...
+            return output_data_type.missing_annotation_default()
+
+    if output_data_type in (AttributeType.LIST, AttributeType.DICT):
+        if not additional:
+            return output_data_type.missing_annotation_default()
+        try:
+            parsed: Any = json.loads(additional)
+        except (json.JSONDecodeError, TypeError):
+            return output_data_type.missing_annotation_default()
```
nikloynes: not a huge fan of a lot of if-else blocks here, could we use a map of some sort instead, or is there a way we could otherwise refactor this to be more extensible and a bit cleaner?

deet/processors/eppi_annotation_converter.py
```diff
@@ -29,6 +29,72 @@
...
+        py_type = output_data_type.to_python_type()
+        if isinstance(parsed, py_type):
+            return cast("EppiRawDataValue", parsed)
+        return output_data_type.missing_annotation_default()
+
+    unsupported = f"Unsupported AttributeType for EPPI mapping: {output_data_type}"
+    raise ValueError(unsupported)
```
nikloynes: i think we could benefit from a specific/custom error here.

nikloynes (approved):
i have to admit that i don't 100% understand what this is doing under the hood, as it would have involved a lot of digging into the code to better understand the context where new code has been added. hence most of my comments are more about understanding/maintainability, and less about the code itself. 


if you're happy with this i'll leave merging it with you.

deet/data_models/processed_gold_standard_annotations.py
sagaruprety: yes. In most datasets people have termed it as report.json. But yeah I'll clarify here again.

deet/data_models/processed_gold_standard_annotations.py
sagaruprety: Kept lazy import and documented why: eppi_annotation_converter imports ProcessedEppiAnnotationData from this module, so a top-level import would circular-import. Happy to extract eppi_output_data_from_eppi_fields to a tiny shared module if we want everything at top level.

deet/evaluators/gold_standard_llm_evaluator.py
sagaruprety: renamed variable to short_snippet_max_len  and added some more documentation around it.

deet/evaluators/gold_standard_llm_evaluator.py
sagaruprety: Renamed to needle/haystack-style names and replaced v/c with readable locals + brief branch comments.

deet/evaluators/gold_standard_llm_evaluator.py
sagaruprety: replaced v/c with readable locals + brief branch comments.

deet/evaluators/gold_standard_llm_evaluator.py
sagaruprety: done

deet/evaluators/gold_standard_llm_evaluator.py
sagaruprety: :  joins multiple EPPI full-text fragments into one CSV cell so we do not rely on commas inside the cell.

deet/evaluators/gold_standard_llm_evaluator.py
sagaruprety: Added an example row to the export_llm_comparison docstring.

deet/processors/eppi_annotation_converter.py
sagaruprety: expanded EPPI terms; explained bool-from-code-row; refactored coercion to match/case for clearer exhaustiveness.

deet/processors/eppi_annotation_converter.py
sagaruprety: Clarified this is a TypeAlias (not a CONST);

deet/processors/eppi_annotation_converter.py
sagaruprety: int(float(additional)) only fails with ValueError, so except ValueError is already the precise branch. A match on exception type wouldn’t add clarity here; we’d still map everything to the same fallback (missing_annotation_default())

deet/processors/eppi_annotation_converter.py
sagaruprety: Refactored: coercion lives in _parse_eppi_integer, _parse_eppi_float, and _parse_eppi_list_or_dict; the main function is a short if dispatcher. I avoided a single dict/callable map because BOOL/STRING don’t share the same parsing signature as the numeric/JSON arms.

deet/processors/eppi_annotation_converter.py
sagaruprety: Done: introduced UnsupportedEppiAttributeTypeError (subclasses ValueError so broad except ValueError still works) with output_data_type attached for logging/handling. Raised from eppi_output_data_from_eppi_fields when the type isn’t mapped.

PR #274, opened 2026-05-21
Use mike to publish multi-version documentation, including development

### LINKED ISSUE #242: Add versioning system for documentation with mike
### Description

We are switching back to keeping main for releases, and merging features into development. This means documentation online will fall behind whatever is in development. We want to be able to share documentation that's in dev with each other, and with other external collaborators, without asking them to build docs themselves.

We should be able to do this with [mike](https://github.com/jimporter/mike) 

### Definition of Done

Github actions ensure that

- [ ] each time we tag a version, deploy the docs to a subfolder in our gh-pages site named `[version]`
- [ ] each time we merge to `development`, deploy the docs to a folder named `development`

### Dependencies

- #207

mcallaghan:
# Pull Request

## Description

- Configure mkdocs.yml to use mike as version provider
- Adjust docs workflow to create a new version of the docs whenever 
  - A version is tagged (update latest)
  - Development is updated (update development)
  
After this, 

- https://destiny-evidence.github.io/data-extraction-evaluation-toolkit/latest will always point to documentation for the latest tagged version
- https://destiny-evidence.github.io/data-extraction-evaluation-toolkit/ will also default to latest
- https://destiny-evidence.github.io/data-extraction-evaluation-toolkit/development will always point to documentation in line with the development branch
- https://destiny-evidence.github.io/data-extraction-evaluation-toolkit/v0.2.0 will always point to documentation for v0.2.0

## Related Issue(s)
Closes #242 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Documentation update

## Pre-Review Checklist

## Testing

Hard to test the workflow, I think we should see what happens when we merge to dev, but have confirmed that

```sh
uv run mike deploy development "Development"
uv run mike serve
```

works as expected.

## Additional Notes

---

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -64,6 +64,4 @@ jobs:
             uv run mike set-default --push latest
           elif [ "${{ github.ref_name }}" = "development" ]; then
             uv run mike deploy --push --update-aliases development
-          elif [ "${{ github.ref_name }}" = "main" ]; then
-            uv run mike deploy --push --update-aliases stable
           fi
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -4,9 +4,13 @@ on:
   push:
     branches:
       - main
+      - development
+    tags:
+      - 'v*'
   pull_request:
     branches:
       - main
+      - development
 
 permissions:
   contents: write
@@ -51,10 +55,15 @@ jobs:
       # first time this workflow step is run.
       - name: Publish documentation on GitHub pages
         if: success() && github.event_name != 'pull_request'
-        uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4
-        with:
-          github_token: ${{  secrets.GITHUB_TOKEN }}
-          publish_dir: site
-          publish_branch: gh-pages
-          user_name: "github-actions[bot]"
-          user_email: "github-actions[bot]@users.noreply.github.com"
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+          if [ "${{ github.ref_type }}" = "tag" ]; then
+            uv run mike deploy --push --update-aliases ${{ github.ref_name }} latest
+            uv run mike set-default --push latest
+          elif [ "${{ github.ref_name }}" = "development" ]; then
+            uv run mike deploy --push --update-aliases development
+          elif [ "${{ github.ref_name }}" = "main" ]; then
+            uv run mike deploy --push --update-aliases stable
+          fi
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -20,22 +20,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout source
-        uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6
-
-      - name: Set up Python
-        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
-          python-version: "3.x"
-          cache: "pip"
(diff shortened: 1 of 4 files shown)


.github/workflows/docs.yml
```diff
@@ -51,10 +55,13 @@ jobs:
...
-          publish_branch: gh-pages
-          user_name: "github-actions[bot]"
-          user_email: "github-actions[bot]@users.noreply.github.com"
+        run: |
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+          if [ "${{ github.ref_type }}" = "tag" ]; then
+            uv run mike deploy --push --update-aliases ${{ github.ref_name }} latest
+            uv run mike set-default --push latest
```
harryjmoss: what happens on pushes to main here (after a PR merge)? Just checking that we're happy with currently excluding that and just using either tags or the dev branch

.github/workflows/docs.yml
```diff
@@ -51,10 +55,13 @@ jobs:
...
+          elif [ "${{ github.ref_name }}" = "development" ]; then
+            uv run mike deploy --push --update-aliases development
+          fi
```
harryjmoss: edit: removed the suggestion that can now be found over in PR #277 that points to the docs/242-multiversion-docs branch

I had to add this to my workflow file in [my fork of the repo](https://github.com/harryjmoss/data-extraction-evaluation-toolkit) to get this working - but the versioning is now working over at [this test github pages site](https://harryjmoss.com/data-extraction-evaluation-toolkit/development/)

Main changes here are that `mike` uses the `gh-pages` branch for its internal versioning and I switched the repo over to deploying from GH actions, rather than from the `gh-pages` branch

Updated this suggestion to make `development` docs point to `latest` - let me know if that's the intended behaviour

harryjmoss: the suggestion might not be formatting properly but effectively I'm suggesting to remove from 
`# Automatically deploy documentation to a GitHub Pages website on pushing to main.` to the end of the file and replace it with [L47 onwards here](https://github.com/harryjmoss/data-extraction-evaluation-toolkit/blob/7856e5f6954a4150d2555436626d2c26716793b5/.github/workflows/docs.yml#L47)

harryjmoss (commented):
I tried this out on a fork of the repo and needed a few tweaks to get this working.

- [Link to the forked repo](https://github.com/harryjmoss/data-extraction-evaluation-toolkit)
- [Link to the built docs](https://harryjmoss.com/data-extraction-evaluation-toolkit/)

Mainly it's stuff in the suggested change above. I also thought we could take this oppportunity to move from the branch-based deployment to deploying via GitHub actions, which also needs:

### Change of build source
Settings -> Pages -> Build and deployment -> Source -> Changed from `Deploy from a branch` to `GitHub Actions`

### Change of permissions for the `gh-pages` Environment

Settings -> Environments -> github-pages -> Add deployment branch or tag rule -> added a rule for tags with a filter of `v*` and added a branch-based rule for `development`

I also had the question above of what we want to do for pushes to `main`, do we ignore docs build & deploys for that and just focus on tagged versions, with `development` pointing to `development`?

.github/workflows/docs.yml
harryjmoss: This currently causes a tagged deployment to point to `latest` in the docs, which I don't think is what we want. Have removed this in the suggestion above

mcallaghan: Our intention in #240 is to always version changes to main by tagging, so this should prevent double builds, as long as we actually do do this!

.github/workflows/docs.yml
mcallaghan: Ah brilliant, smart idea to test this in a fork! If this works great. On latest, see below

.github/workflows/docs.yml
mcallaghan: It was what I intended, in the sense of latest being the latest stable, tagged version, and development being the bleeding edge unreleased version.

Perhaps stable and development would be less ambiguous aliases?

harryjmoss: Ah got it! Will revert that suggestion back to your original implementation in that case. I think `stable` and `development` could work nicely here.

harryjmoss (approved):

PR #275, opened 2026-05-21
Revise DEET sys prompt

kaitlynhair:
Updated to latest version of system prompt

# Pull Request

## Description
Update to system prompt to explicitly reference data types and input/output fields. 

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## Pre-Review Checklist

- [x ] Code passes `ruff` linting

## Testing
Didn't make any changes to main codebase or functionality. 

## Additional Notes

---


deet/prompts/system_prompt.txt
```diff
@@ -1,13 +1,52 @@
...
+
+Your task is to extract information strictly from the provided paper.
+
+For each question, produce a structured response containing the following fields:
+
+Answer: A concise, direct answer to the question. The answer may be;
+
+A Boolean (true/false)
+
+A binary response (yes/no)
```
mcallaghan: Are these different? Should this not just mirror our AttributeTypes?

deet/prompts/system_prompt.txt
```diff
@@ -1,13 +1,52 @@
...
+
+Additional_text: A verbatim excerpt from the paper that directly supports the answer.
+
+Extraction Rules
+
+Use only the content of the provided paper
+
+Do not speculate, infer unstated information, or rely on domain knowledge
+
+Choose simplest valid answer type that fully answers the question (e.g., use Boolean or yes/no when the paper clearly supports it)
```
mcallaghan: Shouldn't we ask it to use the type provided by the Attribute?

deet/prompts/system_prompt.txt
nikloynes: agreed. perhaps there should be a refactor of `data_models.base.AttributeType`, that allows us to define the system prompt for a given `AttributeType` in code (thus ensuring this document stays up to date). it would look a bit like this:

```python
from enum import StrEnum


class AttributeType(StrEnum):
    """Enum of permitted attribute data types."""

    description: str

    def __new__(cls, value: str, description: str):
        obj = str.__new__(cls, value)
        obj._value_ = value
        obj.description = description # we could extend this here with other sysprompt-relevant things.
        return obj

    STRING = ("string", "Text value.")
    INTEGER = ("integer", "Whole number.")
    FLOAT = ("float", "Decimal number.")
    BOOL = ("bool", "True/false value.")
    LIST = ("list", "List of values.")
    DICT = ("dict", "Dictionary/object value.")

    def __str__(self) -> str:
        """Return the string value for JSON serialization."""
        return self.value # this retains existing functionality for access and retrieval of data types
```
we'd then have to dynamically import this into `system_prompt.txt`, e.g. in CI or pre-commit hooks. 


this will
- ensure that the system prompt is complete.
- allows us to easily add new custom, specific attribute types along with prompts in code.

deet/prompts/system_prompt.txt
```diff
@@ -1,13 +1,52 @@
...
 
-IMPORTANT: Be careful to distinguish between the current paper's findings and references to previous literature. When the text refers to other studies (indicated by citation markers like reference numbers [1], author names and years (Smith et al., 2020), etc.), do not use these statements as evidence for the current paper's attributes. Only use statements that directly describe the current paper's methodology.
+You will be given:
 
+an academic paper, provided as a markdown-formatted document file
 
-Paper text is provided inside the JSON under the "context" key.
+a list of questions or attributes for which answers are requested.
+
+Your task is to extract information strictly from the provided paper.
```
nikloynes: maybe not restrict it to `paper`, maybe something like document would be better.

deet/prompts/system_prompt.txt
```diff
@@ -1,13 +1,52 @@
...
+
+A Boolean (true/false)
+
+A binary response (yes/no)
+
+A short string or text span extracted or paraphrased from the paper.
+
+Reasoning: A clear explanation on how the answer was derived from the paper’s content.
+
+Additional_text: A verbatim excerpt from the paper that directly supports the answer.
```
nikloynes: perhaps would make sense to re-inforce that it shouldn't give us anything that doesn't exist

nikloynes (commented):
not going to review remaining diff, perhaps you can merge dev into your feature branch to ensure everything is current, reducing the diff to the sys prompt? 
added a point that we might want to link the specific instructions for a given data type into the code (`data_models.base.AttributeType`).

deet/prompts/system_prompt.txt
kaitlynhair: agreed - will change

PR #278, opened 2026-05-26
Draft: Implement chunked evaluation lifecycle management (add-dev and validate-run)

### LINKED ISSUE #17: split data into train/test splits
we may want to split incoming data into test/train.

mcallaghan:
# Pull Request

### Summary

This PR introduces the data models and CLI workflows for managing document assignment across evaluation phases, following the [Chunked Evaluation Data](https://destiny-evidence.github.io/evaluation-book/index-1/#chunked-evaluation-data) strategy.

The evaluation strategy is designed to cleanly separate data used to develop prompts and adjust model configurations, and data used to evaluate how well prompts and model configurations work.

Since implementing this strategy is not straightforward using the existing CLI (though it could be achieved by creating multiple projects, each for a different evaluation context), this PR adds commands to do this natively in the deet CLI.

### User workflow

Previously, a user of the deet CLI would run `deet experiments evaluate`, and the prompts and model configuration specified would be evaluated against all of the gold standard evaluation data in their project. This immediately burns the whole dataset.

#### 1. A new project `deet projects init`
Now, when the user runs `deet experiments evaluate` in a new project, the command checks how documents are assigned to evaluation splits in the project, finds that the `development` set is empty, and sends the user a message telling them that they need to first assign documents to the development set, and how to do this
 
#### 2. Adding documents to the development set `deet add-dev`

The user runs `deet add-dev -s 5` to add 5 documents selected at random from all unassigned documents to the development set.

#### 3. Do prompt-and-model configuration engineering `deet experiments evaluate`

The user now has documents in their development set with which they can try out different prompts and model configurations. Each time they run `deet experiments evaluate`, the prompts and model configuration specified are used to extract data from documents. The configuration, prompts, and the results of this, are stored in an experiment directory (defined in `deet.data_models.projects.ExperimentArtefacts`. The user keeps doing this until they have performance they are happy with.

#### 4. Validate prompt-and-model configuration on a new validation set `deet experiments validate-run`

The user now runs `deet experiments validate-run -s 5` to add 5 randomly selected unassigned documents to their `validation` set. They are prompted to choose an experiment, whose prompt-and-model configuration is then used to extract data from the validation set. They are presented with metrics showing how well the prompts they lovingly crafted to optimise for their development set perform out of sample. 

They then are prompted to choose between two actions

#### 4a. Accept validation run and proceed to final test

It the user is happy with the validation metrics, they may choose to lock in the prompt-and-model configuration, assign all remaining documents to the test set, and proceed to a final evaluation of their prompt-and-model config. This presents them with their final results, they are now finished.

#### 4b. Reject validation run and return to development

If the validation metrics are poor, the user can conclude that their prompts and model configuration did not generalise well. They clear the documents used for validation, add them into the development set, and proceed to adjust their prompts and configuration, until they are happy with their results. Once they are, they return to step 4.

#### A new module `deet.data_models.evaluation_splits.py` to manage the splitting of documents into different evaluation stages.

```python
class EvaluationSplits(BaseModel):
    """Keeps track of the way documents have been assigned to stages."""

    current_stage: EvaluationStage = EvaluationStage.DEVELOPMENT
    development_ids: list[int] = Field(default_factory=list)
    validation_ids: list[int] = Field(default_factory=list)
    test_ids: list[int] = Field(default_factory=list)
```

EvaluationSplits records the current stage, and a list of IDs assigned to each of the stages. 

It is persisted by being serialised into a location defined with respect to the project directory: `splits.dump_to_json(deet_project.evaluation_splits_path)`. The state of how documents were split for each experiment is also persisted, following how config and prompts are persisted for an experiment: `project_evaluation_splits.dump_to_json(experiment_artefacts.evaluation_splits_snapshot)`

It has the following methods and properties to manage the allocation of IDs to stages, and the retrieval of the right IDs to use at any given time
- `_get_list_for_stage(self, stage: EvaluationStage)` returns a list of IDs for any given stage
- `active_ids` returns the IDs of the currently active stage
- `get_unassigned_ids()` returns all IDs that are currently unassigned to any stage
- `add_to_stage()` adds documents to a stage by randomly sampling `s` IDs from unassigned IDs
- `finalise_test()` adds all remaining unassigned documents to the test set
- `reject_validation()` moves all validation IDs into development 

#### Changes to the CLI `deet.commands.experiments` and `deet.extractors.cli_helpers.py` 

- The helper function `deet.extractors.cli_helpers.run_extraction_pipeline` checks the current state of evaluation splitting, writes this into the experiment directory, and then filters documents for the current stage (using a method added to ProcessedEvaluationData)

https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/5bf97cec7b5f3f45eaaa7ded139cb2608cdf4b6a/deet/extractors/cli_helpers.py#L165-L177 
 
- `deet experiments add-dev` adds unassigned documents to the dev set, and reports how many are still unasssigned
- `deet experiments validate-run` implements stages 4, 4a, and 4b described above

### Todos and future work

- [ ] Add a command to lock in the test set from the beginning
- [ ] *Maybe* implement stratified sampling / or adding specific documents (by ID) to development
- [ ] Deal with interrupted validation runs. At the moment, interrupting a validation run (before deciding to accept or reject) with ctrl-c leaves the state locked at validation. We need a command like resolve-validation. And for that to work we need to track the particular experiment folder being validated.
- [ ] *Possibly* some helper commands to show the current splits and how many documents are still available
- [ ] Documentation
- [ ] Tests

We may also want to discuss how much of the logic lives in the cli, and how much in the EvaluationSplits methods, especially considering use of deet as a python package rather than cli. Most of it is in EvaluationSplits, but CLI commands still manage transitions from stage to stage.

We might also want to consider what happens when a dataset is amended - do we want to check if IDs recorded in EvaluationSplits still match project IDs? I.e. do we want to ensure that the filter documents function returns the correct number of records?

## Related Issue(s)
Closes #17 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Pre-Review Checklist

- [ ] Code passes `mypy` type checking

## Testing
Interactively tested. Awaiting feedback on overall design before writing unit tests.

## Additional Notes

---


deet/extractors/cli_helpers.py
```diff
@@ -26,18 +25,9 @@
 from deet.ui.terminal.wizards import continue_after_key, run_model_wizard
 
 
-def load_config_from_typer_context(
-    typer_context: typer.Context, config_path: Path | None
-) -> DataExtractionConfig:
+def load_or_init_config(config_path: Path | None) -> DataExtractionConfig:
```
mcallaghan, line 31: We already have a guard to make sure commands that require a project are run from a project directory. This is no longer necessary, and avoids us passing typer_context around

deet/extractors/cli_helpers.py
```diff
@@ -57,24 +47,6 @@ def load_config_from_typer_context(
         fail_with_message(f"Config validation error in {config_path}:\n{e}")
 
 
-def init_extraction_run(
```
mcallaghan, line 63: This is superseded by ExperimentArtefacts.create()

deet/extractors/cli_helpers.py
```diff
@@ -96,9 +68,13 @@ def prepare_documents(
     if config.default_context_type == ContextType.FULL_DOCUMENT:
         if linked_document_path.exists():
             notify(f"Loading linked documents from {linked_document_path}")
-            documents = [Document.load(f) for f in linked_document_path.glob("*.json")]
-            if documents:
-                return documents
+            linked_documents = []
```
mcallaghan, line 84: no longer just load everything in the linked_documents folder (this should really be in a separate fix PR)

deet/data_models/enums.py
```diff
@@ -13,3 +13,15 @@ class CustomPromptPopulationMethod(StrEnum):
...
+class EvaluationStrategyName(StrEnum):
+    """
+    A list of allowable names for evaluation strategies.
+
+    Note: make sure that each of these is in the strategy registry in
+    deet.data_models.evaluation_strategies.__init__.py.
+    """
+
+    NONE = auto()
+    DEV_VAL_TEST = auto()
```
nikloynes, line 27: i think we need some more info in comments here on what this should actually be, and what will be achieved by them -- maybe in the docstring. do we need a `NONE` strategy, or could we just pass `None` if we don't want to employ a strategy?

deet/data_models/evaluation_strategies/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+if TYPE_CHECKING:
+    from deet.data_models.project import DeetProject
+
+
+_STRATEGY_REGISTRY: dict[
+    EvaluationStrategyName,
+    Callable[["DeetProject"], BaseEvaluationStrategy],
+] = {
+    EvaluationStrategyName.NONE: NullEvaluationStrategy,
```
nikloynes, line 30: this seems a bit weird to me -- how about just passing none (side-stepping evaluation strats altogether?)

deet/data_models/evaluation_strategies/base.py
```diff
@@ -0,0 +1,117 @@
+"""Base ABC for evaluation strategies and splits objects."""
+
+from __future__ import annotations  # Makes all type annotations lazy strings
```
nikloynes, line 17: are we maintaining compatibility with older versions? afaik we aren't. so we shouldn't need this.

deet/data_models/evaluation_strategies/base.py
```diff
@@ -0,0 +1,117 @@
...
+
+    from deet.data_models.enums import EvaluationStrategyName
+    from deet.data_models.project import DeetProject, ExperimentArtefacts
+
+
+class BaseEvaluationStage(StrEnum):
+    """Defines the stages of an evaluation strategy."""
+
+
+class BaseEvaluationStrategy[SplitsT: BaseSplits](ABC):
```
nikloynes: you can have these (forward) annotations as strings without having the from __future__ block. this will be better if we want to have a more modern python version going forward.

deet/data_models/evaluation_strategies/base.py
```diff
@@ -0,0 +1,117 @@
...
+    """Base class defining methods of an evaluation strategy."""
+
+    name: EvaluationStrategyName
+
+    def __init__(self, project: DeetProject) -> None:
+        """initialise and set splits."""
```
nikloynes: not a problem for me, but i think this should fail pre-commit as not capitalised at start of docstring?

deet/data_models/evaluation_strategies/base.py
```diff
@@ -0,0 +1,117 @@
...
+    def _load_splits(self, project: DeetProject) -> SplitsT:
+        """Return the splits object for this strategy and project."""
+
+    def get_active_ids(self, project: DeetProject) -> list[int]:
+        """Return IDs to run the pipeline on."""
+        return self.splits.active_ids
+
+    def snapshot(self, artefacts: ExperimentArtefacts) -> None:
+        """Persist stategy state alongside an experiment run."""
+        self.splits.dump_to_json(artefacts.evaluation_splits_snapshot)
```
nikloynes: see an older review of mine where i flagged art I facts vs art E facts. not sure which one we went with...

deet/data_models/evaluation_strategies/base.py
```diff
@@ -0,0 +1,117 @@
...
+
+    @abstractmethod
+    def run_splits_wizard(
+        self,
+        project: DeetProject,
+        *,
+        action: str | None = None,
```
nikloynes: what goes in here then? and might it make sense to constrain the range of actions this method can take (e.g. a registry?)

deet/data_models/evaluation_strategies/base.py
```diff
@@ -0,0 +1,117 @@
...
+    """Base object for persisting how document IDs are used for an experiment."""
+
+    _STAGE_FIELD_NAMES: ClassVar[dict[BaseEvaluationStage, str]]
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        """Check whether subclasses have defined _STAGE_FIELD_NAMES."""
+        if "_STAGE_FIELD_NAMES" not in cls.__dict__:
+            missing_names = f"{cls.__name__} must define _STAGE_FIELD_NAMES"
+            raise TypeError(missing_names)
+        super().__init_subclass__(**kwargs)
```
nikloynes: this could do with some more documentation, explaining what's going on. quite confusing to the uninitiated.

deet/data_models/evaluation_strategies/base.py
nikloynes, line 1: i think this file can do with:
- some more explaining/documentation, maybe some inline comments as well
- perhaps some re-jigging (why are the `StageT` etc objects defined after the class in which they're used?)
- perhaps some renaming to slightly more descriptive names (even if longer)

deet/data_models/evaluation_strategies/dev_val_test.py
```diff
@@ -0,0 +1,304 @@
+# ruff: noqa: PLC0415
+"""
+The dynamic dev-val-test splitting strategy.
```
nikloynes, line 3: we maaay have to explicitly tell pytest to exclude this file if we're just running `pytest`.

deet/data_models/evaluation_strategies/dev_val_test.py
```diff
@@ -0,0 +1,304 @@
...
+
+This is described in detail at https://destiny-evidence.github.io/evaluation-book/index-1/#chunked-evaluation-data
+"""
+
+from __future__ import annotations
```
nikloynes, line 8: see previous file.

deet/data_models/evaluation_strategies/dev_val_test.py
```diff
@@ -0,0 +1,304 @@
...
+
+
+class DevValTestSplits(BaseSplits):
+    """Model to record how documents are allocated across dev-val-test splits."""
+
+    _STAGE_FIELD_NAMES: ClassVar[dict[BaseEvaluationStage, str]] = {
+        DevValTestEvaluationStage.DEVELOPMENT: "development_ids",
+        DevValTestEvaluationStage.VALIDATION: "validation_ids",
+        DevValTestEvaluationStage.TEST: "test_ids",
+    }
```
nikloynes, line 55: might the enum, and the fact it's inheriting from an empty enum, be a little overkill here? could we just use a dict?

deet/data_models/evaluation_strategies/dev_val_test.py
```diff
@@ -0,0 +1,304 @@
...
+
+    current_stage: DevValTestEvaluationStage = DevValTestEvaluationStage.DEVELOPMENT
```
nikloynes, line 57: again i find the naming here quite confusing. so we have a devvaltestevaluationstage, which contains dev, val and test? but it doesnt contain evaluation. i think it would help to explain a little bit what's going on, or maybe re-think the naming a little bit?

deet/data_models/evaluation_strategies/dev_val_test.py
```diff
@@ -0,0 +1,304 @@
...
+    validation_ids: list[int] = Field(default_factory=list)
+    test_ids: list[int] = Field(default_factory=list)
+
+    validation_run_id: str | None = None
+
+    @classmethod
+    def load(cls, file_path: Path) -> DevValTestSplits:
+        """Load splits from file."""
+        if not file_path.exists():
+            return cls()
```
nikloynes, line 86: may want to either throw an error here, or at least throw a message, as i imagine not loading something isn't the expected behaviour of `load()`?

deet/data_models/evaluation_strategies/dev_val_test.py
```diff
@@ -0,0 +1,304 @@
...
+            f" This now contains {len(self.splits.development_ids)} documents."
+            f" {len(self.splits.get_unassigned_ids(project_doc_ids))}"
+            " are still unassigned."
+        )
+
+    def _validate_run(
+        self, deet_project: DeetProject, size: int, project_doc_ids: list[int]
+    ) -> None:
+        """Select a past experiment config and eval against a fresh validation set."""
+        from InquirerPy import inquirer
```
nikloynes: just a note now that i'm seeing this, and the `notify` etc. messages -- is this exclusively implemented for CLI? should it not also be implemented as a python API?

deet/data_models/evaluation_strategies/dev_val_test.py
```diff
@@ -0,0 +1,304 @@
...
+        )
+        self.splits.validation_run_id = experiment_artefacts.run_id
+        self.splits.dump_to_json(deet_project.evaluation_splits_path)
+        self.snapshot(experiment_artefacts)
+
+    def _act_on_validation(
+        self, deet_project: DeetProject, project_doc_ids: list[int]
+    ) -> None:
+        """Given validation, choose to return to development or move to testing."""
+        from InquirerPy import inquirer
```
nikloynes, line 271: might make sense to import this stuff at the top of this class, seeing as we're using it in quite a few methods?

deet/data_models/evaluation_strategies/dev_val_test.py
nikloynes, line 1: is there no use-case for having this as non-CLI? if not, perhaps make a little clearer that this is CLI-only.

deet/data_models/evaluation_strategies/null.py
nikloynes, line 1: i see. so you've decided to refactor the app insofar that we now _need_ this `NullStrategy`as otherwise it won't work? and typically the user wouldn't select this, it would just happen as default?

deet/data_models/project.py
```diff
@@ -289,13 +302,49 @@ def process_data(self) -> ProcessedAnnotationData:
...
+    def get_all_doc_ids(self) -> list[int]:
+        """Process the full dataset and return all document IDs."""
+        processed_data = self.process_data()
+        return processed_data.all_doc_ids
+
+    def load_evaluation_strategy(
+        self,
+    ) -> BaseEvaluationStrategy[BaseSplits]:
+        """Load split state."""
+        from deet.data_models.evaluation_strategies import _STRATEGY_REGISTRY
```
nikloynes: wonder if this shouldn't be implicit private, i.e. just `STRATEGY_REGISTRY` given it's quite important?

deet/extractors/cli_helpers.py
nikloynes: maybe stick that in an issue?

mcallaghan: #374

deet/data_models/enums.py
mcallaghan: I think we need to define what happens when you run `deet experiments splits`, and also define how documents are filtered. I suppose we could do this with a few more if clauses, e.g.

https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/3db4b97e2c5f05bbc39efddf3718cd7453e49900/deet/extractors/cli_helpers.py#L191-L194,

but personally I prefer thinking about it as "no strategy is also a strategy". But perhaps we could consider renaming, e.g. to "eval-everything" or "no-splits"

deet/data_models/evaluation_strategies/__init__.py
mcallaghan: as above, I think we'd then need to find all the places that expect a strategy (now and in the future) and put a gate on these. To me it's cleaner in the long run if we say that every project has an evaluation strategy, which defines how to split documents when evaluating. The default strategy however, does not filtering at all, and just prints an informative message when you run `deet experiments splits`

deet/data_models/evaluation_strategies/base.py
mcallaghan: Ah no it's not about backwards compatibility (or at least not backwards of python3.12), but about keeping heavy imports under TYPE_CHECKING and allowing forward references

This is default behaviour (and so this line would not be necessary) from python3.14. So we need it since we support python>=3.12

deet/data_models/evaluation_strategies/base.py
mcallaghan: In principle yes, but as far as I understand it the future import allows us to deal with annotations such that the only code we need to change if we supported python>=3.14 would be to remove this import. If we put the annotations as strings, this would then be superfluous from 3.14

deet/data_models/evaluation_strategies/base.py
mcallaghan: It's actually in the ignore list since the project was set up:

https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blame/3db4b97e2c5f05bbc39efddf3718cd7453e49900/pyproject.toml#L140

But I have capitalised it :)

deet/data_models/evaluation_strategies/base.py
mcallaghan: artefacts = BE, artifacts = AE

deet/data_models/evaluation_strategies/base.py
mcallaghan: Yes this is a good point, and I thought about it. But valid actions are both strategy and stage specific. So we could construct the union of all possible actions - these would then be shown in --help, but at runtime (based on the strategy the project employs), only a subset of these would be valid. I think this would be more confusing.

deet/data_models/evaluation_strategies/base.py
mcallaghan: agreed, addressed in [da2b5d0](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/pull/278/commits/da2b5d033cc34a17691e1e47699429e375ff26ee)

deet/data_models/evaluation_strategies/dev_val_test.py
mcallaghan: pytest is already configured with testpaths = ["tests"] in pyproject.toml

deet/data_models/evaluation_strategies/dev_val_test.py
mcallaghan: I think we would lose type safety in all parts of the code in this strategy that behave dependent on the stage. Would imply introducing a fair few magic strings

deet/data_models/evaluation_strategies/dev_val_test.py
mcallaghan: ah I can see where the confusion comes from, and I've tried to clarify in the docstring.

Each of these members is of type EvaluationStage (inherits from the abstract BaseEvaluationStage).

The name is formed from {strategy name}{base class being inherited from}

The three evaluation stages for the dev-val-test strategy are dev, val, and test.

deet/data_models/evaluation_strategies/dev_val_test.py
mcallaghan: Yes this is unclear. I've renamed to load_or_init, and changed the docstring accordingly.

deet/data_models/evaluation_strategies/dev_val_test.py
mcallaghan: No, by no means. notify doesn't require that this is run from the CLI. To make this clearer, [db67d1d](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/pull/278/commits/db67d1d685374b4cbb3fb37bc9897dfb55b87297) isolates the bits that require interactivity (although there is no reason interactivity can't be part of programmatic use), and makes sure there is a non-interactive way to call all the methods that an API-user might require. I've also made private methods public and improved documentation to make it easier for library-users to use.

deet/data_models/evaluation_strategies/dev_val_test.py
mcallaghan: I think deferring means that imports are only required for methods that actually need them. This keeps things lightweight, especially if we have API users who want to call .validate_run(), say, and don't need Inquirerpy

deet/data_models/evaluation_strategies/dev_val_test.py
mcallaghan: see above

deet/data_models/evaluation_strategies/null.py
mcallaghan: yes exactly. This keeps current behaviour exactly as-is, without leaving ifs all over the place where we check for strategy presence before doing something.

deet/data_models/project.py
mcallaghan: Agreed

PR #303, opened 2026-06-14
feat: add keyword and semantic extractors, and introduce a shared ABC…

### LINKED ISSUE #123: note for re-building the extract_from_documents() for batching

mcallaghan:
# Pull Request

## Description
Introduces keyword extractors as alternative to llms.

For binary attributes, LLM data extraction may be overkill. With large taxonomies, LLMs may even perform worse than other methods.

Alternative zero-shot approaches to classification include simple keyword matching, and matching on semantic similarity to keywords. Both approaches are implemented here, and can be run using the CLI in the same way as the LLMDataExtractor, by setting the `ExtractionMethod` attribute of the `DataExtractionConfig` object. This is not prompted for in the wizard, but can be run using a config yaml file.

To enable this, extraction logic common to any extractor is moved to an abstract base class. All extractors inherit from this. Each extractor now only has to implement its own `extract_from_document()` function.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Additional Notes

This PR still needs 
- [ ] tests
- [ ] to figure out if any UI changes are required in the CLI
- [ ] updates to the documentation

---

### DIFF
--- a/deet/extractors/base_extractor.py
+++ b/deet/extractors/base_extractor.py
@@ -1,0 +1,410 @@
+"""Base data extraction module to be extended with different extraction methods."""
+
+import json
+from abc import ABC, abstractmethod
+from collections.abc import Sequence
+from enum import StrEnum, auto
+from importlib.resources import files
+from pathlib import Path
+from typing import Annotated, Any
+
+import yaml
+from loguru import logger
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    Field,
+    model_validator,
+)
+
+from deet.data_models.base import Attribute
+from deet.data_models.documents import (
+    ContextType,
+    Document,
+    GoldStandardAnnotatedDocument,
+)
+from deet.data_models.extraction import (
+    DocumentExtractionResult,
+    ExtractionRunMetadata,
+    ExtractionRunOutput,
+)
+from deet.data_models.ui_schema import UI
+from deet.exceptions import LitellmModelNotMappedError
+from deet.settings import (
+    DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK,
+    LLMProvider,
+)
+from deet.ui.terminal.render import optional_progress
+from deet.utils.tokenisation import (
+    get_model_max_tokens,
+)
+
+
+def default_system_prompt() -> str:
+    """Get default system prompt included in the package."""
+    return (files("deet.prompts") / "system_prompt.txt").read_text()
+
+
+def _model_string_for_tokenisation(provider: LLMProvider, model: str) -> str:
+    """
+    Build the model string used for tokenisation.
+
+    Must match how ``LLMDataExtractor`` sets ``self.model`` from config.
+    """
+    match provider:
+        case LLMProvider.AZURE:
+            return f"azure/{model}"
+        case LLMProvider.OLLAMA:
+            return f"ollama/{model}"
+        case _:
+            msg = f"Unsupported LLM provider: {provider}"
+            raise ValueError(msg)
+
+
+class PromptConfig(BaseModel):
+    """Configuration for prompts used in data extraction."""
+
+    model_config = ConfigDict()
+
+    system_prompt: str | Path = Field(
+        description="System prompt that defines the task and role",
+        default_factory=default_system_prompt,
+    )
+
+    @model_validator(mode="after")
+    def load_system_prompt_file(self) -> "PromptConfig":
+        """Load system prompt from file if Path provided."""
+        if isinstance(self.system_prompt, Path):
+            if not self.system_prompt.exists():
+                sys_prompt_missing = f"sys prompt {self.system_prompt} not found."
+                raise ValueError(sys_prompt_missing)
+            logger.debug(f"Reading system prompt from {self.system_prompt}")
+            self.system_prompt = self.system_prompt.read_text()
+
+        return self
+
+
+class ExtractionMethod(StrEnum):
+    """Enum of extraction methods."""
+
+    LLM = auto()
+    KEYWORD = auto()
+    SEMANTIC = auto()
+
+
+class DataExtractionConfig(BaseModel):
+    """Configuration for data extraction tasks."""
+
+    model_config = ConfigDict()
+
+    method: ExtractionMethod = Field(
+        default=ExtractionMethod.LLM, description="Extraction Method"
+    )
+
+    # LLM
+    provider: Annotated[
+        LLMProvider, UI(help="Choose from a list of supported LLM providers.")
+    ] = Field(default=LLMProvider.AZURE, description="LLM Provider")
+    model: Annotated[str, UI(help="The name of the LLM model you want to use.")] = (
+        Field(
+            default="gpt-4o-mini",
+            description="LLM model identifier used for completions.",
+        )
+    )
+    temperature: float = Field(
+        default=0.1,
+        description="Sampling temperature for the LLM.",
+        ge=0.0,
+    )
+    max_tokens: Annotated[
+        int | None,
+        UI(
+            help=(
+                "The maximum number of tokens in the LLM response. "
+                "Leave blank for the provider default."
+            )
+        ),
+    ] = Field(
+        default=None,
+        description=(
+            "Maximum number of tokens to generate (Leave blank for provider default)."
+        ),
+    )
+
+    max_context_tokens: Annotated[
+        int | None,
+        UI(
+            help=("Maximum input context length " "(Leave blank for provider default).")
+        ),
+    ] = Field(
+        default=None,
+        description=(
+            "Maximum input context length in tokens (system + attributes + "
+            "document). None = infer from model (litellm registry), else "
+            f"{DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK} via "
+            "DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK. Override to manage costs."
+        ),
+    )
+
+    # Context
+    default_context_type: Annotated[
+        ContextType, UI(help="Where to extract data from.")
+    ] = Field(
+        default=ContextType.FULL_DOCUMENT, description="Type of context to provide"
+    )
+
+    truncate_on_overflow: Annotated[
+        bool,
+        UI(
+            help=(
+                "Select true to truncate documents longer than max_context_tokens. "
+                "This will ensure extraction runs without crashing, but may mean"
+                " some parts of the document are not seen by the LLM."
+            )
+        ),
+    ] = Field(
+        default=False,
+        description=(
+            "When True, automatically truncate context that exceeds "
+            "max_context_tokens. When False (default), raise ValueError."
+        ),
+    )
+
+    # Prompt
+    prompt_config: PromptConfig = Field(
+        default_factory=PromptConfig, description="Prompt configuration"
+    )
+
+    # Output
+    include_reasoning: bool = Field(
+        default=True, description="Include reasoning in output"
+    )
+    include_additional_text: bool = Field(
+        default=True, description="Include additional text/citations in output"
+    )
+
+    @model_validator(mode="after")
+    def populate_max_context_tokens_from_model(self) -> "DataExtractionConfig":
+        """Populate max_context_tokens from model when not set."""
+        if self.max_context_tokens is not None:
+            return self
+        model_str = _model_string_for_tokenisation(self.provider, self.model)
+        try:
+            inferred = get_model_max_tokens(model_str)
+        except LitellmModelNotMappedError:
+            inferred = None
+        if inferred is not None:
+            self.max_context_tokens = inferred
+        else:
+            # Use shared fallback when model max tokens cannot be inferred.
+            self.max_context_tokens = DEFAULT_LLM_MAX_CONTEXT_TOKENS_FALLBACK
+        return self
+
+    @classmethod
+    def from_yaml(cls, path: Path) -> "DataExtractionConfig":
+        """Load config object from a yaml file."""
+        if not path.exists():
+            not_found = f"Config file not found at: {path}"
+            raise FileNotFoundError(not_found)
+
+        return cls.model_validate(yaml.safe_load(path.read_text()))
+
+
+class BaseDataExtractor(ABC):
+    """Abstract Base Class defining common methods for data extractors."""
+
+    def __init__(self, config: DataExtractionConfig) -> None:
+        """Initialise with data extraction config."""
+        self.config = config
+
+    def _prepare_context(
+        self,
+        payload: str,
+        context_type: ContextType | None = None,
+    ) -> str:
+        """Prepare context based on context type."""
+        ctx = (
+            context_type
+            if context_type is not None
+            else self.config.default_context_type
+        )
+        logger.debug(f"Using context type: {ctx}")
+        if ctx == ContextType.FULL_DOCUMENT:
+            context = payload
+            logger.debug(f"Using full document context (length: {len(str(context))})")
+        elif ctx == ContextType.ABSTRACT_ONLY:
+            context = payload
+            logger.debug(f"Using abstract context (length: {len(str(context))})")
+        elif ctx == ContextType.RAG_SNIPPETS:
+            rag_not_impl = "rag-snippets context type is not implemented."
+            raise NotImplementedError(rag_not_impl)
+        elif ctx == ContextType.CUSTOM:
+            custom_not_impl = "custom context type is not implemented."
+            raise NotImplementedError(custom_not_impl)
+        else:
+            other_not_allowed = f"{ctx} context type is not allowed."
+            raise ValueError(other_not_allowed)
+
+        if isinstance(context, list):
+            logger.debug(f"Converting list context to string (items: {len(context)})")
+            context = " ".join(context)
+
+        return context
+
+    def extract_from_documents(  # noqa: PLR0913
+        self,
+        attributes: list[Attribute],
+        documents: Sequence[Document],
+        filter_attribute_ids: list[int] | None = None,
+        output_file: Path | None = None,
+        context_type: ContextType | None = None,
+        prompt_outfile: Path | None = None,
+        *,
+        show_progress: bool = False,
+    ) -> ExtractionRunOutput:
+        """
+        Extract data from all documents.
+
+        Loops over documents and extracts data using list of attributes.
+
+        Args:
+            attributes: List of attributes to extract.
+            documents: Sequence of Document instances (required).
+            filter_attribute_ids: Optional list of attribute IDs to filter by.
+            output_file: Optional path to save combined results JSON.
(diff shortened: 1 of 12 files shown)


deet/extractors/base_extractor.py
```diff
@@ -0,0 +1,545 @@
...
+            logger.debug(f"Using abstract context (length: {len(str(context))})")
+        elif ctx == ContextType.RAG_SNIPPETS:
+            rag_not_impl = "rag-snippets context type is not implemented."
+            raise NotImplementedError(rag_not_impl)
+        elif ctx == ContextType.CUSTOM:
+            custom_not_impl = "custom context type is not implemented."
+            raise NotImplementedError(custom_not_impl)
+        else:
+            other_not_allowed = f"{ctx} context type is not allowed."
+            raise ValueError(other_not_allowed)
```
nikloynes: we used the `match/case` construct earlier. do we want to replicate that approach here as well?

deet/extractors/base_extractor.py
```diff
@@ -0,0 +1,545 @@
...
+        ) as iterable_documents:
+            for document in iterable_documents:
+                logger.info(f"Processing document: {document.name}")
+
+                try:
+                    if context_type == ContextType.ABSTRACT_ONLY:
+                        document.set_abstract_context()
+                    elif context_type == ContextType.FULL_DOCUMENT:
+                        document.context = document.safe_parsed_document.text
+
```
nikloynes: could we refactor this a little bit -- it risks getting quite a bit cumbersome if we implement n different options here. perhaps we can have a registry at the top of the class and map to that here?

deet/extractors/base_extractor.py
```diff
@@ -0,0 +1,545 @@
...
+                    parsing = parsing_by_doc.get(doc_id_str, DocumentParsingStats())
+                    per_document[doc_id_str] = PerDocumentExtractionStats(
+                        input_tokens=result.input_tokens,
+                        output_tokens=result.output_tokens,
+                        parsing_seconds=parsing.parsing_seconds,
+                        parsing_skipped=parsing.parsing_skipped,
+                        llm_call_seconds=result.llm_call_seconds,
+                    )
+                    total_input_tokens += result.input_tokens
+                    total_output_tokens += result.output_tokens
```
nikloynes: is this still relevant when we're doing keyword extraction? should it perhaps be in the llm-specific one?

deet/extractors/base_extractor.py
```diff
@@ -0,0 +1,545 @@
...
+                    if result.total_cost_usd is not None:
+                        total_cost = (total_cost or 0.0) + result.total_cost_usd
+
+                except NoAbstractError as e:
+                    logger.warning(f"Skipping {document.name}: {e}")
+                except Exception as e:  # noqa: BLE001
+                    logger.error(f"Failed to process {document.name}: {e}")
+                    logger.debug("Error details", exc_info=True)
```
nikloynes: i am always quite hesitant with these. presumably we want to be able to continue with our extraction even if something fails somewhere. but: we should have enough knowledge at this stage to know what the errors are that interrupt our run, no? can we specify more of these?

deet/extractors/base_extractor.py
```diff
@@ -0,0 +1,545 @@
...
+
+        run_metadata = ExtractionRunMetadata(
+            model=self.config.model,
+            total_input_tokens=total_input_tokens,
+            total_output_tokens=total_output_tokens,
```
nikloynes: see the previous message. presumably this will be `None` for a keyword-only run. 
maybe we should think about what all `None` can mean in this context, and perhaps have an enum instead which has zero ambiguity re the tokens.

deet/extractors/base_extractor.py
```diff
@@ -0,0 +1,545 @@
...
+    def from_yaml(cls, path: Path) -> "DataExtractionConfig":
+        """Load config object from a yaml file."""
+        if not path.exists():
+            not_found = f"Config file not found at: {path}"
+            raise FileNotFoundError(not_found)
+
+        return cls.model_validate(yaml.safe_load(path.read_text()))
+
+
+class BaseDataExtractor(ABC):
```
nikloynes: i find the naming slightly confusing. we use pydantic, and hence `Base`model quite a lot. noe we're calling this a `Base` data extractor, but it's not actually pydantic. might be worth renaming to `AbstractDataExtractor`, or something like that?

deet/extractors/keyword/base_keyword_extractor.py
```diff
@@ -0,0 +1,27 @@
+"""Shared classes for keyword-type extractors."""
+
+from deet.data_models.base import Attribute, AttributeType
+from deet.extractors.base_extractor import BaseDataExtractor
+
+
+class BaseKeywordDataExtractor(BaseDataExtractor):
+    """
+    ABC for keyword data extractors, handling tasks common to them.
```
nikloynes: is this still an ABC, or is this a concrete implementation of an ABC?

deet/extractors/keyword/raw_keyword_extractor.py
```diff
@@ -0,0 +1,79 @@
...
+
+class RawKeywordDataExtractor(BaseKeywordDataExtractor):
+    """
+    String-matching implementation of a keyword extractor.
+
+    Extracts the presence of attributes where phrases in attribute prompts
+    match the document content.
+    """
+
+    SNIPPET_WINDOW = 80
```
nikloynes: how come we set this to 80? why not an arg with default?

deet/extractors/keyword/raw_keyword_extractor.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _snippet_around(self, context: str, idx: int, phrase_len: int) -> str:
+        """Return a window of the document around the first match of phrase."""
+        start = max(0, idx - self.SNIPPET_WINDOW)
```
nikloynes: somewhat confusing as to how snippet window and phrase length differ. intuitively, they should be the same thing.

deet/extractors/keyword/semantic_keyword_extractor.py
```diff
@@ -0,0 +1,145 @@
...
+from deet.extractors.base_extractor import DataExtractionConfig
+from deet.extractors.keyword.base_keyword_extractor import BaseKeywordDataExtractor
+
+
+class SemanticKeywordDataExtractor(BaseKeywordDataExtractor):
+    """
+    Keyword extractor matching attribute phrases to document sentences.
+
+    Uses sentence-transformer embeddings and cosine similarity rather than
+    exact string matching, so semantically related wording still matches.
```
nikloynes: will be interesting to do some comparison of LLMs vs this approach.

deet/extractors/keyword/semantic_keyword_extractor.py
```diff
@@ -0,0 +1,145 @@
...
+    def __init__(self, config: DataExtractionConfig) -> None:
+        """Initialise, set the similarity threshold, and load the model."""
+        super().__init__(config)
+        self.similarity_threshold: float = config.semantic_similarity_threshold
+        self.model = SentenceTransformer(config.model)
+
+    def _split_into_sentences(self, text: str) -> list[str]:
+        """Split text into sentences on terminal punctuation followed by a space."""
+        sentences = re.split(r"(?<=[.!?]) +", text)
+        return [s.strip() for s in sentences if s.strip()]
```
nikloynes: noting that this will struggle with tables and/or bulleted lists. might consider having each of those as 'sentences' as well.

deet/extractors/keyword/semantic_keyword_extractor.py
```diff
@@ -0,0 +1,145 @@
...
+        """Embed a list of texts into a (len(texts), dim) array of vectors."""
+        return self.model.encode(texts, show_progress_bar=False, convert_to_numpy=True)
+
+    def _phrases_per_attribute(self, attributes: list[Attribute]) -> list[list[str]]:
+        """
+        Return the keyword phrases for each attribute, in order.
+
+        A keyword attribute with no usable prompt has nothing to match on, so an
+        empty result is presumed to be an error.
+
```
nikloynes: maybe you can add an example here in the docstring, i find this quite hard to wrap my head around but think an example would make it clearer.

deet/extractors/keyword/semantic_keyword_extractor.py
```diff
@@ -0,0 +1,145 @@
...
+        group_sizes = [len(phrases) for phrases in phrases_per_attribute]
+        split_points = np.cumsum(group_sizes)[:-1]
+        return np.split(similarity_matrix, split_points, axis=0)
+
+    def _best_sentence_match(
+        self,
+        similarities: np.ndarray,
+        phrases: list[str],
+        doc_chunks: list[str],
+    ) -> tuple[float, str, str]:
```
nikloynes: could consider a pydantic model for the output of this method

deet/extractors/llm_data_extractor.py
nikloynes: awesome how much we save here with this refactor

docs/concepts/data-extraction.md
```diff
@@ -169,3 +169,64 @@ gold_standard_document = GoldStandardAnnotatedDocument(
...
 
 represents the fact that the document named "document 1" had been labelled as relevant by a human.
+
+## Extraction methods
+
+Once we have documents and attributes, `deet` needs to work out the value of
+each attribute for each document. How it does this is controlled by the `method`
+field on [`deet.extractors.base_extractor.DataExtractionConfig`](../reference/api.md#deet.extractors.base_extractor.DataExtractionConfig),
+which takes one of the [`deet.extractors.base_extractor.ExtractionMethod`](../reference/api.md#deet.extractors.base_extractor.ExtractionMethod)s.
+The method we choose also changes how an attribute's `prompt` is interpreted.
```
nikloynes: i think this - how the prompt is interpreted - could perhaps also be reflected in in-line documentation (docstrings)

docs/concepts/data-extraction.md
```diff
@@ -169,3 +169,64 @@ gold_standard_document = GoldStandardAnnotatedDocument(
...
+`prompt` is a question or instruction written in natural language, for example
+"Does the article discuss the effects of climate change on human health?". This
+is the only method that produces values for every attribute type, and the LLM is
+chosen with the `provider` and `model` fields.
+
+### Keyword
+
+The keyword method does no LLM call at all: it simply checks whether the words we
+are looking for appear in the document. Here the `prompt` is not a question but a
+list of keyphrases, separated by semicolons (`;`). An attribute is marked `True`
```
nikloynes: in the example below we've actually separated it with semicolon + space

tests/unit/test_cli.py
```diff
@@ -502,7 +502,7 @@ def test_extract_happy_path(tmp_path):
     with (
         patch("deet.data_models.project.DeetProject.load") as mock_loader,
         patch("deet.extractors.cli_helpers.run_model_wizard") as mock_wizard,
-        patch("deet.extractors.cli_helpers.LLMDataExtractor") as mock_extractor_cls,
+        patch("deet.extractors.cli_helpers.get_data_extractor") as mock_get_extractor,
```
nikloynes: do we test this new get_data_extractor method somewhere?

tests/unit/test_extractor_contract.py
```diff
@@ -0,0 +1,60 @@
...
+from deet.extractors.base_extractor import DataExtractionConfig, ExtractionMethod
+from deet.extractors.extractor_registry import get_data_extractor
+
+
+def _attr(attribute_id: int) -> Attribute:
+    return Attribute(
+        attribute_id=attribute_id,
+        attribute_label=f"Attribute {attribute_id}",
+        output_data_type=AttributeType.BOOL,
+    )
```
nikloynes: should maybe be a fixture

tests/unit/test_extractor_contract.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.fixture(params=list(ExtractionMethod), ids=lambda m: m.value)
+def extractor(request):
+    """Each registered extractor, constructed via the registry."""
+    config = DataExtractionConfig(method=request.param)
+    with patch(
+        "deet.extractors.keyword.semantic_keyword_extractor.SentenceTransformer"
+    ):
+        return get_data_extractor(config=config)
```
nikloynes: we want to have a generic extractor fixture, but we're patching in semantic keyword extractor sentence transformer. is that right? will we just ignore if not relevant?

tests/unit/test_extractor_register.py
```diff
@@ -0,0 +1,26 @@
...
+from deet.extractors.extractor_registry import extractor_mapping, get_data_extractor
+
+
+@pytest.mark.parametrize("extraction_method", list(ExtractionMethod))
+def test_extraction_methods_return_extractor(extraction_method):
+    """Test that each member of ExtractionMethod returns a valid extractor."""
+    config = DataExtractionConfig(method=extraction_method)
+    with patch(
+        "deet.extractors.keyword.semantic_keyword_extractor.SentenceTransformer"
+    ):
```
nikloynes: see previous file

tests/unit/test_keyword_extractor.py
```diff
@@ -0,0 +1,107 @@
...
+    return DataExtractionConfig(
+        method=ExtractionMethod.KEYWORD,
+        default_context_type=ContextType.FULL_DOCUMENT,
+    )
+
+
+@pytest.fixture
+def extractor(config) -> RawKeywordDataExtractor:
+    """Keyword extractor under test."""
+    return RawKeywordDataExtractor(config=config)
```
nikloynes: i feel like these names could perhaps be a bit more specific, as `config` and `extractor` i would normally expect in conftest.

nikloynes (commented):
big piece of work. looks great overall. some not super pressing feedback in comments.

deet/extractors/base_extractor.py
mcallaghan: this is not new code, but stuff I moved over - but the answer is probably yes. I'd defer to another issue

deet/extractors/base_extractor.py
mcallaghan: see above. Yes this whole block is not as nice as it could be imo

deet/extractors/base_extractor.py
mcallaghan: Not exactly relevant, but it's important that the output is always in the same shape. With the keyword extractor, tokens will just be 0

deet/extractors/base_extractor.py
mcallaghan: again, old code moved over, but let's make a new issue for this. We saw in the demo this morning, that these errors get hidden in all the mass of logs produced

deet/extractors/base_extractor.py
mcallaghan: it will be zero, which I think is correct. The number of tokens was 0

deet/extractors/base_extractor.py
mcallaghan: Hm, I'm following the pattern across the repo, so I think we should keep this here, but let's make an issue to resolve and standardise naming conventions for abstract / base classes

deet/extractors/keyword/base_keyword_extractor.py
mcallaghan: It is in fact still an ABC since it doesn't implement the concrete method `extract_from_document`, but I have clarified the docstring to make this clearer

deet/extractors/keyword/raw_keyword_extractor.py
mcallaghan: good point - arg with default

deet/extractors/keyword/raw_keyword_extractor.py
mcallaghan: explained further in updated docstring

deet/extractors/keyword/semantic_keyword_extractor.py
mcallaghan: some results here: https://destiny-evidence.github.io/impact-case-1/taxonomy/

deet/extractors/keyword/semantic_keyword_extractor.py
mcallaghan: yes, I'd postpone this for future work :)

deet/extractors/keyword/semantic_keyword_extractor.py
mcallaghan: yes, done

deet/extractors/keyword/semantic_keyword_extractor.py
mcallaghan: yes, might improve readability, but I think this has some performance overhead, and there's no validation of messy data to do

deet/extractors/llm_data_extractor.py
mcallaghan: :), yes imo this makes it much easier to read what the extractor actually does

docs/concepts/data-extraction.md
mcallaghan: yes, I've added this on prompts to the Attribute docstring

docs/concepts/data-extraction.md
mcallaghan: implicitly, yes - and I've clarified the implied separation

tests/unit/test_cli.py
mcallaghan: I think this is covered by test_extractor_register.py

tests/unit/test_extractor_contract.py
mcallaghan: done, and also re-used this in a couple of other places for DRYness

tests/unit/test_extractor_contract.py
mcallaghan: yes, the reason is that initing a SentenceTransformer tries to load an embedding model, and we just need to make sure it doesn't have that.

tests/unit/test_keyword_extractor.py
mcallaghan: yes, renamed and this is clearer. Not put into conftest as these are specific to each extractor

mcallaghan:
> big piece of work. looks great overall. some not super pressing feedback in comments.

Thanks! Think I've addressed everything here now

deet/extractors/base_extractor.py
nikloynes: #405

deet/extractors/base_extractor.py
nikloynes: #406

nikloynes (approved):

PR #309, opened 2026-06-15
feat: more intuitive commands for project creation and use

mcallaghan:
# Pull Request

## Description
Feedback on the UI detailed confusion that `deet project init` did not create a new directory (this is intended, it turns the current directory into a deet project).

This PR introduces a new command `deet project new`, with a required argument for the project name, that creates a new directory and initialises the project in there.

It also changes `deet project init` to no longer prompt for the project name. Instead, this is taken from the current directory.

Doing this required changing how we deal with paths to project data, which are entered relative to the current directory, and stored relative to project root (this had always been the same place). Now that these are not necessarily the same place, paths are entered relative to the current directory, and then anchored to the target directory with `DeetProject.anchor_to()`.

> This PR also implements a recursive search upwards from the current directory to find a project.yaml file. This means that deet can be run not only inside a project directory, but also from any subdirectory of a deet project directory.

EDIT: This is no longer the case, as this requires changes to how settings are loaded. Was not worth it and is not required.

If users want to store their data (pdfs and gold standard annotated data) in the same directory as their deet project, then they should still use deet project init. 

I think that `deet project init` will suit most users better, as they won't have to `cd` after creating a project (which for terminal-shy users won't feel intuitive). Changing the users directory within the command would require spawning an additional terminal process, which I think would end up more confusing.

Documentation of both commands is improved to clarify how to deal with directories.

This also fixes the missing autocomplete on pdf path entry, and swaps the order of data format and path

## Related Issue(s)
Closes #292 Closes #294, Closes #293

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## Pre-Review Checklist

## Testing
Interactive testing, new unit tests

## Additional Notes

---


deet/data_models/project.py
```diff
@@ -53,12 +51,22 @@ class DeetProject(BaseModel):
         str,
         UI(
             help="Give your project a name. This will help you to identify it later",
-            valid="Must be at least 2 characters",
+            valid="Must be at least 1 character",
```
nikloynes, line 54: what's the motivation behind changing from 2 to 1?

deet/data_models/project.py
```diff
@@ -70,27 +78,18 @@ class DeetProject(BaseModel):
             instructions="press Tab to autocomplete, '/' to go to next directory",
             valid="Must be a valid .csv or .json path",
```
nikloynes, line 79: does it validate this?

deet/data_models/project.py
```diff
@@ -136,6 +135,29 @@ def config_path(self) -> Path:
...
+    @property
+    def gold_standard_data_abspath(self) -> Path:
+        """
+        Return a usable path to the gold-standard data.
+
+        ``gold_standard_data_path`` is stored relative to the project root (and
+        never persisted as an absolute path, so ``project.yaml`` stays portable).
+        This joins it with the root only for I/O.
+        """
+        return self.root / self.gold_standard_data_path
```
nikloynes, line 147: `Path` has built-in method for this:
```python
In [2]: from pathlib import Path

In [3]: test = Path('test.txt')

In [4]: test
Out[4]: PosixPath('test.txt')

In [5]: test.absolute()
Out[5]: PosixPath('/Users/nikloynes/test/test.txt')
```

deet/data_models/project.py
```diff
@@ -151,29 +173,64 @@ def check_suffix(cls, value: Path) -> Path:
             raise ValueError(unsupported_ext)
         return value
 
-    @field_validator("pdf_dir", mode="after")
+    @field_validator("pdf_dir", mode="before")
     @classmethod
-    def _process_pdf_dir(cls, value: Path) -> Path | None:
-        """Parse empty string to None (not cwd), otherwise return path."""
-        if value == "" or value is None:
+    def _process_pdf_dir(cls, value: object) -> object | None:
```
nikloynes, line 178: what's the `object` doing here vs a `Path`?

deet/data_models/project.py
```diff
@@ -151,29 +173,64 @@ def check_suffix(cls, value: Path) -> Path:
...
+        correct when ``root`` differs from that directory (as with
+        ``deet project new``). The stored values remain relative, never absolute.
+        """
+        source = source_dir or Path.cwd()
+        self.gold_standard_data_path = (
+            source / self.gold_standard_data_path
+        ).relative_to(root, walk_up=True)
+        if self.pdf_dir is not None:
+            self.pdf_dir = (source / self.pdf_dir).relative_to(root, walk_up=True)
+        self._root = root
```
nikloynes, line 199: i might be wrong, but might we be able to refactor this a little using the `.absolute()` and other methods in `Path`?

deet/data_models/project.py
```diff
@@ -151,29 +173,64 @@ def check_suffix(cls, value: Path) -> Path:
...
+            )
+            raise FileNotFoundError(missing_pdfs)
+
     def setup(self) -> None:
         """
         Set a project up.
 
         Create directory structure, process gold-standard data, and create
             prompt csv and link map
         """
```
nikloynes, line 222: so will this method work for both the `init` and `new` commands? if so, should it take args? or have we fixed this with the instance attributes being more flexible?

deet/data_models/project.py
```diff
@@ -185,8 +242,15 @@ def setup(self) -> None:
...
-    def dump_to_yaml(self, target: Path = PROJECT_FILE) -> None:
-        """Write a minimal ``project.yaml`` file to save project options."""
+    def dump_to_yaml(self, target: Path | None = None) -> None:
+        """
+        Write a minimal ``project.yaml`` file to save project options.
+
+        Written to the project root by default. Resource paths are stored as their
+        relative values, never resolved to absolute, so the file stays portable.
+        """
+        target = target if target is not None else self.root / PROJECT_FILE
```
nikloynes: should perhaps ensure that target has a `.yml` or `.yaml` file ending

deet/data_models/project.py
```diff
@@ -202,20 +266,33 @@ def export_config_template(self) -> None:
...
+        data = yaml.safe_load(project_file.read_text())
+        project = cls.model_validate(data["project"])
+        project._root = root  # noqa: SLF001
+        return project
 
     @classmethod
-    def exists(cls) -> bool:
-        """Check if project exists in current directory."""
-        return PROJECT_FILE.exists()
+    def exists(cls, project_dir: Path | None = None) -> bool:
```
nikloynes: wouldn't it make sense to validate presence using the project data model (wherein we're presumably defining what needs to be present for a project to exist/be valid?

deet/scripts/commands/project.py
```diff
@@ -10,115 +10,79 @@
...
-    project.setup()
+@app.command()
+def new(
+    *,
+    name: Annotated[
+        str | None, typer.Option(help="Project name (prompted if omitted).")
+    ] = None,
+    data_path: DataPathOption = None,
+    data_type: DataFormatOption = SupportedImportFormat.EPPI_JSON,
+    pdf_dir: PdfDirOption = None,
```
nikloynes, line 64: sidenote, but we always use pdf interchangeably with markdown, right?

deet/scripts/project_utils.py
```diff
@@ -0,0 +1,177 @@
...
+    init_md = render_template("project/init")
+    console.print(info_panel(init_md, title=":speedboat: project set-up"))
+    continue_after_key()
+
+    project = run_model_wizard(DeetProject, prefill={"name": name})
+    project.anchor_to(root)
+    project.setup()
+
+    console.clear()
+    configure_env_md = render_template("project/configure_env.md")
```
nikloynes, line 44: replcae with constant instead of magic value?

deet/scripts/project_utils.py
```diff
@@ -0,0 +1,177 @@
...
+    console.print(
+        info_panel(
+            render_template(
+                "project/success.md", project=project, new_directory=str(new_directory)
+            )
+        )
+    )
+
+
+def guard_overwrite(target_dir: Path, *, force: bool, interactive: bool) -> None:
```
nikloynes, line 62: really smooth! will bookmark this for future projects!

deet/scripts/project_utils.py
```diff
@@ -0,0 +1,177 @@
...
+        raise ValueError(no_ui)
+    ui_help = ui.help + (
+        ". The name you enter will be standardised and used to create a directory"
+    )
+    console.clear()
+    console.print(wizard_field_help("name", ui_help))
+    return str(inquire_pydantic_field(DeetProject, "name", info, ui))
+
+
+# Shared options, so `init` and `new` accept identical headless arguments.
```
nikloynes: what does 'headless' mean in this context?

deet/scripts/typer_context.py
```diff
@@ -43,8 +43,14 @@ def wrapper(*args: P.args, **kwargs: P.kwargs) -> R:
 
         if not typer_context or not typer_context.obj or not typer_context.obj.project:
             no_project = (
-                "This command must be run from a directory that contains a project"
-                " create one by running `deet project init`"
+                "This command must be run from a directory that contains a project."
+                "\n\nCreate one in this directory by running `deet project init`,"
+                "\n\nor create a project in a"
+                " new directory by running"
+                " `deet project new`."
```
nikloynes, line 50: this is slightly confusing -- wasn't the change that we need to supply a directory (project) name with one of the commands? so that should probably be represented in the message to the user.

deet/ui/terminal/templates/welcome.md
```diff
@@ -19,11 +19,17 @@ deet experiments --help
...
 to learn how to extract data from the documents in your project
 {% else %}
-To start using deet, set up a project by running
+To start using deet, set up a project in the current directory by running
 
 ```sh
 deet project init
 ```
 
+If you prefer to create a new directory, you can do this by running
```
nikloynes, line 28: same here, don't we need to supply the dir/project name?

deet/ui/terminal/wizards.py
```diff
@@ -44,7 +44,7 @@ def can_handle(self, field_info: FieldInfo) -> bool:
 
     def execute(self, widget_args: dict[str, Any], field_info: FieldInfo) -> str:
         """Execute an inquirer.select prompt."""
```
nikloynes: this is nit-picky, but maybe the term 'prompt' should be reserved entirely for LLM-relevant prompts.

docs/gen/project_init.tape
nikloynes, line 1: cool - what does this do?

deet/data_models/project.py
mcallaghan: In one path, the name is taken directly from the current directory, so any valid directory should be a project name, hence the change

deet/data_models/project.py
mcallaghan: yes, and this is the message that appears if it doesn't pass validation

deet/data_models/project.py
mcallaghan: Yes, but absolute resolves relative to the current working directory. At the moment, deet must be run from a project directory, so cwd will always be project.root, but if we ever change this (as I tried in earlier commits, before deciding this is overly complicated), to e.g. run from a sub-directory, then this will break. self.root is already absolute, so this property does deliver an absolute path.

deet/data_models/project.py
mcallaghan: this is because it's not yet a Path. In fact, we don't yet know what it is, which is why we need to validate it and return a Path (or None). In the previous snippet we did if value == "", which would never be the case for a Path

deet/data_models/project.py
mcallaghan: We're using the .relative_to method. We don't want to produce an absolute path, but just make sure relative paths are relative to the project dir, not the cwd (this is different if we do `deet project new`, which creates a project in a new directory)

deet/scripts/project_utils.py
mcallaghan: without the wizards, I'll change this to "non-wizard"

deet/scripts/project_utils.py
mcallaghan: Good idea, but IMO it's only used once, so I'm not sure it's warranted, would just be an extra line of code - this would also make it inconsistent with other uses of render_template

deet/scripts/typer_context.py
mcallaghan: ah, based on @kaitlynhair 's feedback, the name for the new directory can be elicited interactively. So deet project new prompts for a name, and deet project new --name new-project would create a new project called new-project. I interpreted Kaitlyn's preference to do this interactively as intended default behaviour, so would leave this instruction in here. BUT, I do need to update the PR description, which I do now

deet/data_models/project.py
mcallaghan: yes this works the same for both, because setup calls `self.process_data()`, which now uses `self.gold_standard_data_abspath`, instead of the path relative to the project root (which we want to be the one stored and persisted for portability), it doesn't matter that setup is being called from a working directory that is not the directory where the project lives.

deet/data_models/project.py
mcallaghan: I've got rid of the target parameter. This is in fact never used, and it is clearer to make sure that the project file is always dumped to self.root / PROJECT_FILE.

deet/scripts/commands/project.py
mcallaghan: yes good point, probably we want to make this clear across deet and docs, but I would do this later

deet/data_models/project.py
mcallaghan: I think here it's worth separating whether something exists, from whether it is valid. This existence check is mainly to guard against overwriting, and we should do this even for invalid projects (e.g. because the yaml file is corrupted, or paths to data no longer exist). Validity is already covered by `.load()`.

deet/ui/terminal/templates/welcome.md
mcallaghan: see above

deet/data_models/project.py
mcallaghan: But I actually dropped it. It's no longer used since we check for project existence not just in cwd. I.e. the old guard before overwriting used this, but now it has to check also in a new directory.

docs/gen/project_init.tape
mcallaghan: records the process of project creation, and creates a gif which we show [here](https://destiny-evidence.github.io/data-extraction-evaluation-toolkit/development/setup/quickstart/)

deet/ui/terminal/wizards.py
mcallaghan: agreed, have edited this now

deet/scripts/typer_context.py
nikloynes: can't it accept a name in the command immediately or have it added interactively if omitted?

mcallaghan: yes, exactly, this is what it does, or do you mean it should be `deet project new new-project`? so a typer argument rather than an option? Can also change to this I think (seem to remember it should be possible to do an optional argument with default None, but I will double check).

PR #314, opened 2026-06-20
Docs/296 documentation fixes

mcallaghan:
# Pull Request

## Description

User feedback pointed out some inconsistencies and points of unclarity in the documentation (both help messages and in docs).

This PR addresses each of the points in #296, commit-by-commit. It would make sense to review commit-wise.

All changes touch only documentation, except for the removal of unimplemented ContextTypes (or ContextTypes we don't want users to select). This is because the wizard for populating DataExtractionConfig uses the enum to present users with choices.

A less drastic way to achieve this would probably be to make only some of these selectable, possibly by subclassing ContextType with only those that we want users to be able to success (full text or abstract), and using this subclass to type context_type in DataExtractionConfig. However it made sense to me to remove those that we intend to implement but haven't yet, and mark them with TODO. Happy to try this if preferable.

## Related Issue(s)
Closes #296 

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Documentation update

## Pre-Review Checklist

## Testing
Documentation changes only. Edited tests that check that specific unimplemented context types that were present in the enum raised an error.

## Additional Notes

---

### DIFF
--- a/deet/scripts/commands/project.py
+++ b/deet/scripts/commands/project.py
@@ -217,6 +217,17 @@ def link(typer_context: typer.Context) -> None:
         )
         linked_document.save(file_path)
 
+    # TODO: Nice message explaining what happened
+    console.print(
+        info_panel(
+            render_template(
+                "project/linked",
+                linked_documents=linked_documents,
+                documents=processed_annotation_data.documents,
+            )
+        )
+    )
+
 
 @app.command()
 def test_llm_config(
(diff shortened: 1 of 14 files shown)


deet/scripts/commands/project.py
```diff
@@ -182,6 +182,17 @@ def link(typer_context: typer.Context) -> None:
         )
         linked_document.save(file_path)
 
+    # TODO: Nice message explaining what happened
+    console.print(
```
harryjmoss, line 209: `console` undefined here (plus `info_panel`, `render_template`), I'll push a patch to import these

harryjmoss: Fixed in ad4f852e98fe78ac969795b57f42431a6c06cf68

PR #319, opened 2026-06-22
feat(experiments): persist run metadata (cost, tokens) to output dir

sagaruprety:
Closes #306. 

The deet experiment output directory now stores the cost of each experiment run as a dedicated, discoverable artefact. The run cost/token data was already computed (ExtractionRunMetadata) but was only buried inside llm_annotations.json. This change writes it to its own run_metadata.json alongside the other output files, giving us a clean place for cost and future run-level metadata.

Changes
**ExperimentArtefacts (deet/data_models/project.py):** added a run_metadata property → run_metadata.json.

**run_extraction_pipeline (deet/extractors/cli_helpers.py):** writes run_output.metadata to run_metadata.json 
after extraction. Applies to both experiments evaluate and experiments predict, since both go through this function.

**Tests:** added a test asserting run_metadata.json is created with the expected model/token/cost fields, and updated the existing extract happy-path test mock.

New output file looks like this: 

```
{
  "model": "azure/gpt-4o-mini",
  "total_input_tokens": 50973,
  "total_output_tokens": 1114,
  "total_cost_usd": 0.009146,
  "per_document_tokens": {
    "126730555": {
      "input_tokens": 15714,
      "output_tokens": 320
    },
    "126730554": {
      "input_tokens": 17475,
      "output_tokens": 448
    },
    "126730556": {
      "input_tokens": 17784,
      "output_tokens": 346
    }
  },
  "timestamp": "2026-06-22T12:35:05.864139Z"
}
```

### DIFF
--- a/deet/data_models/project.py
+++ b/deet/data_models/project.py
@@ -254,3 +254,8 @@ class ExperimentArtefacts:
     def llm_annotation_csv(self) -> Path:
         """Return location of csv containing llm extractions."""
         return self.base_dir / "llm_annotations.csv"
+
+    @property
+    def run_metadata(self) -> Path:
+        """Return location of json capturing run metadata (cost, tokens, model)."""
+        return self.base_dir / "run_metadata.json"
--- a/deet/extractors/cli_helpers.py
+++ b/deet/extractors/cli_helpers.py
@@ -210,4 +210,10 @@ def run_extraction_pipeline(
         encoding="utf-8",
     )
 
+    experiment_artefacts.run_metadata.write_text(
+        run_output.metadata.model_dump_json(indent=2),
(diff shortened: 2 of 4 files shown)


deet/data_models/project.py
```diff
@@ -254,3 +254,8 @@ def llm_annotations(self) -> Path:
     def llm_annotation_csv(self) -> Path:
         """Return location of csv containing llm extractions."""
         return self.base_dir / "llm_annotations.csv"
+
+    @property
+    def run_metadata(self) -> Path:
```
nikloynes, line 259: probably best to call it run_metadata_path?

nikloynes (approved):
looks good!

PR #321, opened 2026-06-25
Fix/99 dynamic llm response schema

### LINKED ISSUE #99: Response format passed to LLM does not ensure desired response
### Description

When an LLM is prompted, we define a schema for the [response](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/683fa91d408708d8ac99da3da39b1c4c340b6d47/deet/extractors/llm_data_extractor.py#L354). This is defined [here](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/683fa91d408708d8ac99da3da39b1c4c340b6d47/deet/data_models/base.py#L301). 

It means we get a list of [LLMAnnotationResponse](https://github.com/destiny-evidence/data-extraction-evaluation-toolkit/blob/683fa91d408708d8ac99da3da39b1c4c340b6d47/deet/data_models/base.py#L270) objects, but there is nothing to ensure that the list is of the length we require, or that the type of `output_data` is of the correct type. This has a couple of issues:

- specifying output_data of type Any, means that (at least with Ollama), the schema is not seen as valid, we need to define the types here
- Having type Any means we may get a response of the wrong type
- Specifying a list does not ensure we get a response for each of the attributes we are interested in. We may get a response for only some attributes (observed already), or we may get a response for different attributes entirely

We can fix this by generating the schema dynamically according to the list of attributes we are interested in. This may involve changing the LLMResponseSchema model

### Definition of Done

- [ ] The schema we pass to an LLM ensures that we get a response for all attributes we are interested in, and that the output_data is the correct type for every attribute

### Dependencies

- #231

sagaruprety:
## Summary
- Replace loose `LLMResponseSchema` (`output_data: Any`, list of annotations) with a dynamically generated per-attribute Pydantic model keyed by `attribute_<id>`.
- Enforce `extra="forbid"` and typed `output_data` so the LLM must return exactly one correctly typed response per requested attribute.
- Update system prompt, extractor parsing, and unit tests.

## Test plan
- [x] `pytest tests/unit/test_base.py tests/unit/test_llm_data_extractor.py` — 106 passed
- [x] `ruff` + `mypy` + pre-commit hooks pass
- [x] Manual CLI A/B on PIK_HIC (5 attributes, gemma3:1b):
  - `development`: 1–2/5 attributes returned, mixed bool/string types, hallucinated attribute id 6
  - `fix/99`: 5/5 attributes every doc, all bool, no hallucinations

Closes #99

### DIFF
--- a/deet/extractors/llm_data_extractor.py
+++ b/deet/extractors/llm_data_extractor.py
@@ -790,8 +790,11 @@ class LLMDataExtractor:
             attribute_id = attribute_id_from_response_key(field_name)
             attribute = attributes_by_id.get(attribute_id)
             if attribute is None:
-                logger.warning(f"No attribute found for ID: {attribute_id}")
-                continue
+                msg = (
+                    f"No attribute found for ID: {attribute_id}. "
+                    "The dynamic response model and attribute list are out of sync."
+                )
+                raise ValueError(msg)
 
             attribute_response = getattr(validated_response, field_name)
             additional_text = (
--- a/deet/extractors/llm_data_extractor.py
+++ b/deet/extractors/llm_data_extractor.py
@@ -22,7 +22,8 @@ from deet.data_models.base import (
     Attribute,
     GoldStandardAnnotation,
     LLMInputSchema,
-    LLMResponseSchema,
+    attribute_id_from_response_key,
+    build_llm_response_model,
 )
 from deet.data_models.documents import (
     ContextType,
@@ -133,9 +134,7 @@ class DataExtractionConfig(BaseModel):
 
     max_context_tokens: Annotated[
         int | None,
-        UI(
-            help=("Maximum input context length " "(Leave blank for provider default).")
-        ),
+        UI(help=("Maximum input context length (Leave blank for provider default).")),
     ] = Field(
         default=None,
         description=(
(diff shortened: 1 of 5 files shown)


deet/extractors/llm_data_extractor.py
```diff
@@ -762,48 +780,40 @@ def _parse_llm_response(
...
-            )
-
-            if not attribute:
-                logger.warning(
-                    f"No attribute found for ID: {llm_annotation.attribute_id}"
-                )
+        for field_name in response_model.model_fields:
+            attribute_id = attribute_id_from_response_key(field_name)
+            attribute = attributes_by_id.get(attribute_id)
+            if attribute is None:
```
mcallaghan: This *shouldn't* happen anymore. Perhaps we should rather raise an error, as it would indicate that something fundamental has gone wrong with the dynamic model creation or validation

mcallaghan (changes requested):
This is incredibly useful, thanks Sagar! 

Before I get deeper in, I came across one larger point that I think is probably worth addressing first.

If we have 5 attributes all with string output, we don't actually need to define a model for each of these. The only difference will be the name. 

It would be more parsimonious to generate a schema per attribute type - this could even be a method of AttributeType. Each attribute would then require the relevant schema.

